PR #5, opened 2020-11-19
Adding protrusions.

luigi-rosso:
image


neurowave:
That was such a rad comic! Also, ouch.

PR #12, opened 2021-01-16
Add Typescript Definitions

GrandSchtroumpf:
Following https://feedback.rive.app/111
- Add typescript definition in publish folder based on `bindings.cpp` file
- Update `package.json` to expose typings
- Update readme with typescript example


mjohnsullivan:
LGTM

mjohnsullivan:
We'll test this out a bit first before merging.

GrandSchtroumpf:
Sure. The types are based on the bindings.cpp file which seems to be a ahead of the current version en rive-canvas. Some properties of LinearAnimation types are not implemented in the current rive-canvas version for example.

GrandSchtroumpf:
I've added some getter for readonly properties on the classes. Let me know if I missed some.

publish/README.md
```diff
@@ -19,3 +19,17 @@ As with all npm packages, there's a freely available CDN via unpkg.com:
...
+    
+## In a Typescript Project
+
+```typescript
+import Rive, { File } from 'rive-canvas';
+
+async function loadRivFile(filePath: string): Promise<File> {
+  const req = new Request(filePath);
+  const loadRive = Rive({ locateFile: (file) => '/node_modules/rive-canvas/'+file, });
+  const loadFile = fetch(req).then((res) => res.arrayBuffer()).then(buf => new Uint8Array(buf));
```
umberto-sonnino, line 31: Since this is an `async` function, why not make this fully async by using `async/await` here, and return a `File` instead of a `Promise`?

publish/types.d.ts
```diff
@@ -0,0 +1,264 @@
...
+  get fps(): number;
+  get workStart(): number;
+  get workEnd(): number;
+  get loopValue(): number;
+  get speed(): number;
+  apply(artboard: Artboard, time: number, mix: number): void;
+}
+
+export declare class LinearAnimationInstance {
+  /** Time of the animation from 0 to 1 */
```
umberto-sonnino, line 159: `time` starts at 0 but is not bound to be at max 1

publish/types.d.ts
```diff
@@ -0,0 +1,264 @@
...
+  time: number;
+  constructor(animation: LinearAnimation);
+  /** increment the animation to a time in sec */
+  advance(sec: number): any;
+  /**
+   * apply animation on the artboard
+   * @param artboard the Artboard on which apply the frame
+   * @param mix 0-1 the strengh of the animaiton in the animation mix
```
umberto-sonnino, line 167: A couple of typos here..! =)

umberto-sonnino (commented):
Thanks for putting this together!

I've highlighted a couple of things that could be changed.
Also, I saw a few inconsistencies in the formatting, so it'd be great if this could be run through a formatter - even the default one in VSCode works!

publish/README.md
GrandSchtroumpf: `loadRive` & `loadFile` can be performed in concurrency as they don't depend on each other. 
The `await` should be done on the `Promise.all` to be the most efficient in timing.

GrandSchtroumpf:
@umberto-sonnino Thanks for the review. I clean it up a bit as requested.

GrandSchtroumpf:
After verification, it's possible to use the OffscreenCanvas to run rive in a Worker : 
```typescript
const el = document.getElemetById('rive');
const canvas = ('OffscreenCanvas' in window) ? el.transferControlToOffscreen() : el;
const ctx = this.canvas.getContext('2d');
rive.CanvasRenderer(ctx); // Types is ok now
```
Let me know if there are other change requests

umberto-sonnino:
LGTM! 
I think it just needs a rebase to not conflict with the README changes from the other PR we merged in a couple of days ago

umberto-sonnino (approved):

PR #15, opened 2021-01-24
Add support for Webworker

### LINKED ISSUE #14: SVGMatrix is deprecated and prevents Rive to run WebWorker
As described in MDN documentation, [SVGMatrix](https://developer.mozilla.org/en-US/docs/Web/API/SVGMatrix) is deprecated in favor of [DOMMatrix](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix).

Using the `DOMMatrix` constructor instead of [svg.createMatrix](https://github.com/rive-app/rive-wasm/blob/master/js/renderer.js#L1) would also help Rive run in a webwoker as `document` is undefined in a WebWorker.

GrandSchtroumpf:
fix #14

## Description
This PR add support for running Rive inside a WebWorker. There are many advantages for running an Canvas animation inside a WebWorker, this [article](https://developers.google.com/web/updates/2018/08/offscreen-canvas) is a good resource for example.

## Content
- Remove dependency on `document` to make it work with `OffscreenCanvas`
- Use `DOMMatrix` instead of deprecated `SVGMatrix`
- Update Readme with example for WebWorker.

I put a full working example inside the Readme.md. I'm happy to move it elsewhere if you have a better place in mind 😊.

### DIFF
--- a/js/renderer.js
+++ b/js/renderer.js
@@ -1,7 +1,5 @@
-var svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
-
 function makeMatrix(m2d) {
-    var m = svg.createSVGMatrix();
+    const m = new DOMMatrix();
     m.a = m2d.xx;
     m.b = m2d.xy;
     m.c = m2d.yx;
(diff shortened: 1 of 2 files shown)


mjohnsullivan:
This is great, and probably time we embraced the DOMMatrix. LGTM.

PR #28, opened 2021-02-06
Rive js types

GrandSchtroumpf:
Add types for `rive-js`.
Update package.json accordingly
Fix small typo: 
`if (loopValue < 0 || loopValue > loopTypes.length)` should be `if (loopValue < 0 || loopValue >= loopTypes.length)`, else value 3 is a valid value.

### DIFF
--- a/js/src/rive.js
+++ b/js/src/rive.js
@@ -65,7 +65,7 @@ var Rive=function(){var e="undefined"!=typeof document&&document.currentScript?d
     * Loop event constructor
     */
     var LoopEvent = function ({animationName, loopValue}) {
-        if (loopValue < 0 || loopValue > loopTypes.length) {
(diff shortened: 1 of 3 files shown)


js/src/rive.js
```diff
@@ -65,7 +65,7 @@ var Rive=function(){var e="undefined"!=typeof document&&document.currentScript?d
     * Loop event constructor
     */
     var LoopEvent = function ({animationName, loopValue}) {
-        if (loopValue < 0 || loopValue > loopTypes.length) {
+        if (loopValue < 0 || loopValue >= loopTypes.length) {
```
mjohnsullivan, line 68: Nice catch!

mjohnsullivan:
Thank you adding those types.

We're in the process of experimenting with moving the code over to Typescript. I'd welcome feedback once it's ready.

PR #40, opened 2021-02-24
Add testing

mjohnsullivan:
Adds Jest testing, support for TypeScript, and proper runtime bundling.

Updates Wasm to publish both minified and unminified source for Rive Wasm and Rive pure


luigi-rosso (approved):
LGTM!

PR #43, opened 2021-02-27
Bug fixes

mjohnsullivan:
Fixes bug with how pausing named animations work and how playing animations are reported.


mjohnsullivan:
Also fixes examples to use the right file.

PR #47, opened 2021-03-12
WIP on Typescript

GrandSchtroumpf:
### Goal of the Draft PR
This is a draft on moving the Rive file into Typescript.
The goal of this draft was to get more acquainted with the code. I figured out it could be useful to be shared. Happy to close it if this is already an ongoing work on your side

### The current stat
There is still many errors in to file, and missing types. 
I tried to keep the code as close as possible to the rive.js is currently, I mainly : 
- transformed Rive into a class
- add "private" keyword for method with "_"
- add types
- change self by this

### Future work
There is still some work I would like to do to deliver an error-free file: 
- import missing types from rive-canvas somehow.
- create a `getEvent(name)` method to fix missing typings.

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -6,6 +6,7 @@ import {
   playbackStates,
   RuntimeLoader,
   LoopEvent,
+  Animation
 } from './utils';
 
 // Lets webpack know to copy the Wasm file to the dist folder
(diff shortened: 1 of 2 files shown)


GrandSchtroumpf:
- For typing reasons I moved each event property into a record and created a private `getEvents(name)` methods.
- I put `any` for types coming from `rive-canvas` for now.
- Remaining errors come from import/export

mjohnsullivan:
Thank you for starting to do this, but we're in the middle of doing much the same task at the moment.

I'd suggest you pause this for the moment while we finish the first pass and then we can review?

GrandSchtroumpf:
sure, no problem :)

mjohnsullivan:
And thank you again for starting this; it's great to have more input on our Typescript conversion.

PR #50, opened 2021-03-16
Converts the Rive object and prototype functions to Typescript

mjohnsullivan:


mjohnsullivan:
Complete rewrite in Typescript; support for Next testing; Webpack now packing for both a web script embed build and an ES6 build.

mjohnsullivan:
Docs still need updating; will get that in shortly.

mjohnsullivan:
@GrandSchtroumpf I can't add you as a reviewer for some reason, but would love feedback.

GrandSchtroumpf:
Awesome work! I'll take a deeper look at this on Monday if it's not too late for you.

mjohnsullivan:
README and CHANGELOG updated.

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
+import Runtime from '../../wasm/publish/rive.js';
```
GrandSchtroumpf, line 1: I think rive-canvas should be a dependency of rive-js in package.json as they are two different libs on npm. This would avoid including `rive-canvas` in the bundle size of `rive-js` and let app developer favorite builder do the magic. 
But I don't know how it would work if your use unpkg directly...

Also, you can use [webpack path plugin](https://www.npmjs.com/package/tsconfig-paths-webpack-plugin) to manage dependency between your projects more easily.

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+
+// #region LoopEvent
+
+// Loop types. The index of the type is the value that comes from Wasm
+export const loopTypes: string[] = ['oneShot', 'loop', 'pingPong'];
```
GrandSchtroumpf, line 6: Small typescript trick, as this array should read only, you can do 
```typescript
export const loopTypes = ['oneShot', 'loop', 'pingPong'] as const;
export type LoopType = typeof loopTypes[number];
```
And you derive a string literal type out of your array. Which reduce the potential typo errors

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+
+/// Loop events are returned through onloop callbacks
+interface LoopEvent {
+  animation: string;
+  name: string;
+  type: number;
+}
+
+// Creates a new LoopEvent
+export let createLoopEvent = (animation: string, loopValue: number): LoopEvent => {
```
GrandSchtroumpf, line 16: I don't think you want to mutate this variable. I would recommend to export it as a regular function, or as a const.
Also I think using string literal will provide a better devX than a number for the loopValue :
```typescript
function createLoopEvent(animation: string, loopType: LoopType)
```

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+// #endregion
+
+// Maps the playback state to the Wasm enum values
+export const playbackStates = { 'play': 0, 'pause': 1, 'stop': 2 };
+
+// #region layout
+
+// Fit options for the canvas
+export enum Fit {
+  Cover     = 'cover',
```
GrandSchtroumpf, line 36: I usually don't recommend to use enum when it's point to a string value. String literal have lighter footprint (it's only a type so it's not complied in the end) and provide a better devX most of the time :
```typescript
type Fit = 'cover' | 'contain' | ...;
```

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+export class Layout {
+  public fit: Fit;
+  public alignment: Alignment;
+  public minX: number;
+  public minY: number;
+  public maxX: number;
+  public maxY: number;
+
+  constructor(
+    fit: Fit = Fit.None,
```
GrandSchtroumpf, line 78: When parameters of the constructor become properties of the class you can save a good amount of code like that :
```typescript
class Layout {
  constructor(
    public fit: Fit = 'none',
  ) {}
}
```
Typescript will automatically manage the boilerplate of defining the property and initializing it 🤩

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+    this.maxY = maxY;
+  }
+
+  // Alternative constructor to build a Layout from an interface/object
+  static new({fit, alignment, minX, minY, maxX, maxY}: LayoutParameters) : Layout {
+    return new Layout(fit, alignment, minX, minY, maxX, maxY);
+  }
+
+  // Returns fit for the Wasm runtime format
+  public runtimeFit(rive: any): any {
```
GrandSchtroumpf, line 99: I would recommend to string type as much as possible the public methods. Of course this is a long term process, but this will reduce enormously the amount of required documentation.

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+    switch (this.fit) {
```
GrandSchtroumpf, line 100: Switch map are quite slow in JS and increase the bundle size as very verbose. You can do :
```typescript
return rive.Fit[this.fit] || 'none';
```

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+}
+
+// #endregion
+
+// #region runtime
+
+// Callback type when looking for a runtime instance
+export type RuntimeCallback = (rive: typeof Runtime) => void;
+
+// Runtime singleton; use getInstance to provide a callback that returns the
```
GrandSchtroumpf, line 152: I would recommend JSdoc format comments for exported API to let the editor display information to the developer. This also will reduce the amount of required documentation :
```typescript
/**
 * Runtime singleton...
 */
```

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+      RuntimeLoader.runtime = rive;
+      // Fire all the callbacks
+      while (RuntimeLoader.callBackQueue.length > 0) {
+        RuntimeLoader.callBackQueue.shift()(RuntimeLoader.runtime);
+      }
+    });
+  }
+
+  // Provides a runtime instance via a callback
+  public static getInstance(callback: RuntimeCallback): void {
```
GrandSchtroumpf, line 193: I would recommend using a Promise based API from the start instead of callback:
- it enables syntax sugar like `async` / `await`
- it provides a better integration in JS API with `Promise.all` and `Promise.allSettle` for example.
- it enables performance optimization in the browser (the method won't be blocking in the current tick as it would be flagged as asynchrone)
- it prevent you to maintain a callback API for retro compatibility reasons.

For example :
```typescript
callbackBased(cb) {
  this.queue.push(cb)
}
promiseBased() {
  return new Promise((res, rej) => this.queue.push(res));
}
```

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+  Load      = 'load',
+  LoadError = 'loaderror',
+  Play      = 'play',
+  Pause     = 'pause',
+  Stop      = 'stop',
+  Loop      = 'loop',
+}
+
+// Event fired by Rive
+export interface Event {
```
GrandSchtroumpf, line 257: `Event` is already a type of the DOM API, overriding DOM API types might lead to unexpected side effect (wrong errors in the compiler). If it's exposed I would recommend more specific name like `RiveEvent`.

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+      }
+    }
+  }
+}
+
+// #region Rive
+
+// Interface for the Rive static method contructor
+export interface RiveParameters {
+  canvas: HTMLCanvasElement, // canvas is required
```
GrandSchtroumpf, line 346: Technically it can also be an `OffscreenCanvas` that runs in a webworker no ?

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+      // Add the default animation
+      const animation = this._artboard.animationAt(0);
+      const instance = new this._rive.LinearAnimationInstance(animation);
+      this._animations.push(new Animation(animation, instance));
+    }
+  }
+
+  // Tracks the last timestamp at which the animation was rendered. Used only in
+  // draw().
+  private lastRenderTime: number;
```
GrandSchtroumpf, line 648: This is consider a good practice to define all the properties before the constructor

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+      type: type,
+      callback: callback, 
+    });
+  }
+
+}
+
+
+// Loads Rive data from a URI via fetch.
+let loadRiveFile = async (src: string): Promise<ArrayBuffer> => {
```
GrandSchtroumpf, line 931: Same as above, I don't think you want to reassign this function. Using `const` or `function` might be better. Also no need to hard type return value if Typescript can infer it for you:
```typescript
// Typescript should be able to know from the code that the return value is Promise<ArrayBuffer>
function loadRiveFile(src: string) {
  const req = new Request(src);
  return fetch(req).then(res => res.arrayBuffer());
}
```

js/package.json
```diff
@@ -24,14 +25,18 @@
...
   "license": "MIT",
   "files": [
-    "dist/rive.js",
+    "dist/rive.min.js",
+    "dist/rive.min.js.map",
+    "dist/rive.dev.js",
+    "dist/rive.dev.js.map",
     "dist/rive.wasm",
-    "src/rive.d.ts"
+    "src/rive.ts"
```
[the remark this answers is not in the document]
mjohnsullivan: So I should include the definition file, rather than the source file? Need to se how to get the compiler to create the types file.

js/src/rive.ts
mjohnsullivan: Good catch; I'm going to move this to a string enum, which will give us consistency with the other string array types.

js/src/rive.ts
mjohnsullivan: Cool; I'll look into that.

js/src/rive.ts
mjohnsullivan: This one is for internal use only, and the type's coming from the Wasm Rive object; struggled with properly typing these.

js/src/rive.ts
mjohnsullivan: Good point on switches; I'll see how better to handle these once types are sorted out.

js/src/rive.ts
mjohnsullivan: I'll read up on JSdoc

js/src/rive.ts
mjohnsullivan: I've implemented both here to see how both work for users; I might get rid of the non-Promise one.

js/src/rive.ts
```diff
@@ -0,0 +1,965 @@
...
+
+// #region animations
+
+// Wraps animations and instances from the runtime and keeps track of playback
+// state
+export class Animation {  
+  public loopCount: number = 0;
+  public paused: boolean = false;
+
+  constructor(private animation: any, public readonly instance: any) {}
```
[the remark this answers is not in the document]
mjohnsullivan: I haven't been using the type definitions from rive-canvas yet; I need to get those in.

js/src/rive.ts
mjohnsullivan: Good point; I'll see about a better name.

js/src/rive.ts
mjohnsullivan: Not sure; I'll check.

js/src/rive.ts
mjohnsullivan: Left-over from porting the js; I'll tidy that up.

js/src/rive.ts
mjohnsullivan: I tend to write out the types to make it clear what's being returned. Let me make that const.

mjohnsullivan:
There's a bunch of great comments here, which we're going to address in later PRs; this is getting merged now in preparation for creating the 0.6.0 build and then the 0.7.0 build for the updated Wasm and runtime formats.

js/package.json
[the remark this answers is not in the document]
GrandSchtroumpf: Yes, typescript will take care of binding the js and d.ts output when the lib is going to be imported.
To generate it, you need to add the flag `declaration` in the tsconfig:
```
"compilerOptions": {
  "declaration": true,
  ...
}
```
here is the doc: https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

PR #53, opened 2021-03-24
publish types in rive-canvas again

GrandSchtroumpf:
Typescript throw compilations errors for `ng-rive` if types are not exported by `rive-wasm`.


mjohnsullivan (approved):
LGTM

mjohnsullivan:
Good catch!

PR #71, opened 2021-04-12
State machine support

mjohnsullivan:
- Adds functionality to make retrieving a Rive file's contents easier
- Adds StateMachine and StateMachineInput classes
- Merges animation and state machine playback management


luigi-rosso (approved):
LGTM, awesome!

PR #76, opened 2021-04-15
Generate ES6 module version of rive-canvas

mjohnsullivan:
- Updates the build scripts to output both an ES6 module and no module versions of rive-canvas
- Wasm binary is now embedded in the (m)js file. To fetch it, use:

```js
Rive({
    locateFile: (file) => 'file://' + file,
}).then((rive) => {};
```
- Added a few simple example HTML files, mostly there for testing


mjohnsullivan:
FYI: @GrandSchtroumpf

PR #79, opened 2021-04-15
Updates canvas build to generate wasm file

mjohnsullivan:
- Updates to latest wasm
- Changes to packaging to fix importing issues in frameworks such as vue and parcel
- Adds parcel example
- rive-canvas is now integrated through package.json
- rive-canvas typescript definitions now used
- wasm binary is now embedded in the js package; no more behind the scenes fetch to retrieve it


mjohnsullivan:
fixes #78

PR #93, opened 2021-04-30
Export StateMachineInput class

avivian:
The CommunityShot page gets all the stateMachineInputs to determine what to put in the inputs section, so it's super helpful to have this type as well!


mjohnsullivan (approved):
LGTM

PR #94, opened 2021-05-08
Implement scrubbing feature

nilskj:
A common feature in website animations is to make the animation play/scrub with browser scroll. This is a proposal to add this feature to the JS Runtime. 

- Fix some minor spelling
- Implement a mechanism for scrubbing, conditionally check this in draw method to draw a single frame.
- Add example of scroll-scrubbing in examples/_frameworks/parcel_example_web


nilskj:
> It doesn't allow to mix looping animations AND some "scrubbing" animations ; is there a way to allow both ?

Great point! By not interfering with the render loop (not changing renderer lastTime, rerunning the loop, this is now possible. See the example where the 'idle' animation keeps running while 'bouncing' and 'windshield_wipers' are scrubbed at different times.

mjohnsullivan:
We're currently using node v14 LTS, which has a different package lock structure to latest. this is causing the package locks to differ greatly. Would you be happy to regenerate them using v14, or for me to do it?

nilskj:
> We're currently using node v14 LTS, which has a different package lock structure to latest. this is causing the package locks to differ greatly. Would you be happy to regenerate them using v14, or for me to do it?

Sure thing, I tried regen with LTS v14.17, is that any better?

mjohnsullivan:
That's great; I had a few suggestions about merging the scrubbing into the main draw loop. I forked your fork and have a PR up with those in there (yep, it's a fork of a fork).

Let me know what you think.

mjohnsullivan (approved):
LGTM

mjohnsullivan:
Thank you @nilskj for making this PR; this is really useful to have in the api.

PR #96, opened 2021-05-13
Path Flattening tools

luigi-rosso:
- Adds a "tools" build option.
- Adds bindings for path flattening and pure-js build of tools library.


mjohnsullivan (approved):
LGTM

PR #105, opened 2021-05-26
Adds new module that does not bundle wasm

mjohnsullivan:
Bundling the wasm in js makes it easy to use the rive runtime module, but at the cost of size, as the bundled wasm is base64 encoded, making it about 25% larger.

This adds a lean version of the js runtime with no wasm bundling; this requires the wasm binary to be separately hosted and the url provided to the js module before a Rive object is instantiated with ```RuntimeLoader.setWasmUrl```.

See the [parcel module example](https://github.com/rive-app/rive-wasm/tree/master/js/examples/_frameworks/parcel_example_module) for how this is used.


luigi-rosso (approved):
LGTM! Thanks for fixing the StateMachine advance changes too!

Issue #114, opened 2021-06-25
rive-js ci


luigi-rosso:
Improve Rive-JS CI such that tests run on push and merge to master deploys to NPM.

https://www.notion.so/Rive-JS-and-Rive-WASM-4c6d80180c1a4698acbed9007b8ab776

PR #115, opened 2021-06-25
Improve CI

luigi-rosso:
improves #114


luigi-rosso:
I'm merging this to check that the publishing to NPM works.

PR #118, opened 2021-06-28
adding rive-cpp check to rive-wasm

### LINKED ISSUE #117: rive-wasm report on falling behind rive-cpp

mjtalbot:
fixes #117 

same as what we have in rive-android and rive-ios (btw rive-wasm is out of date so we should get alerts for this... )

### DIFF
--- a/.github/workflows/submodule_check.yaml
+++ b/.github/workflows/submodule_check.yaml
@@ -1,0 +1,24 @@
+name: Check rive-wasm sub modules
+
+on:
+  # lets check it when we push to master (we've just done something in android, its a good time to see about this)
+  push:
+    branches:
+      - master
+  
+  schedule:
+    # 8am UTC every day
+    - cron:  '0 8 * * *'
+
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Run cpp version check 
+        env:
+            SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        run: sh ./.github/scripts/rive_cpp_check.sh
--- a/.github/scripts/rive_cpp_check.sh
+++ b/.github/scripts/rive_cpp_check.sh
@@ -1,0 +1,20 @@
+set -e
+
+# GET hash of CPP MASTER.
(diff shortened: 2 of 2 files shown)


avivian (approved):
We could create our own action for this step: https://docs.github.com/en/actions/creating-actions. Means our changes are centralised and we don't need to update loads of repos when we make changes. Though I don't know if there's much we'd want to change about the script

umberto-sonnino:
> We could create our own action for this step: https://docs.github.com/en/actions/creating-actions. Means our changes are centralised and we don't need to update loads of repos when we make changes. Though I don't know if there's much we'd want to change about the script

That's pretty cool!

.github/workflows/submodule_check.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - master
+  
+  schedule:
+    # 8am UTC every day
+    - cron:  '0 8 * * *'
+
+
+jobs:
+  build:
+    runs-on: macos-latest
```
umberto-sonnino, line 16: Just noticing that this is running on macos, and I think we've got much less free compute time for these machines (i.e. they're much more expensive). Any reason to run them on mac rather than linux?

mjtalbot: naa.. no reason whatsoever. isnt it free though?

mjtalbot: just reading this it *sounds* free https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions (doesnt mean that we need to be using macos-latest of course :P )

umberto-sonnino: I think we have a number of free compute minutes each month, and after a while they run out.
In fact, [macOS has a higher multiplier minutes](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions) which might make them run out faster.

mjtalbot: but the first sentence? 

```
About billing for GitHub Actions
GitHub Actions usage is free for both public repositories and self-hosted runners.
```

umberto-sonnino: Oh, you're right! Nvmd then! Free 🍻

PR #125, opened 2021-07-20
Updating rive-cpp.

luigi-rosso:


luigi-rosso:
Super simple, just updated the submodule and had to fix a small issue with the build-js script.

umberto-sonnino (approved):

PR #126, opened 2021-07-20
Update rive cpp

luigi-rosso:


luigi-rosso:
This is a fix for what's causing problems with rive-react.  rive_canvas.js was going to src instead of dist which wouldn't be included in the npm module.

PR #133, opened 2021-08-12
Clear whole canvas on each render to allow for noclip artboards

avivian:


mjtalbot (approved):
lgtm. gotta love clearing the canvas!

luigi-rosso (commented):
LGTM, nice catch.

luigi-rosso:
In the future we could expose Artboard.clip and only do the full clear if that is true, if it proves to be an issue.

PR #137, opened 2021-08-31
Exposing some more transform properties.

luigi-rosso:
Exposing some more transform properties for anything inheriting from TransformComponent. Also exposes the ability to get the renderer alignment so it can be used to compute screen to world/artboard conversions (nice for getting mouse/touch coordinates into artboard space).


mjtalbot (approved):
lgtm. a bunch of this access stuff could probably have a very basic test for it

luigi-rosso:
That'd be interesting to get setup in this repo...a test case that uses the resulting JS/WASM and runs some JS tests.

PR #139, opened 2021-09-07
Realign on every draw

avivian:
Fixes ghosting when the canvas aspect ratio does not match the artboard aspect ratio.

I've also added some examples for no clip artboards, to make testing this fast in the future


js/src/rive.ts
```diff
@@ -1232,45 +1228,28 @@ export class Rive {
     }
   }
 
-  /**
-   * Clears the canvas
-   */
-  private clearCanvas(): void {
```
mjtalbot, line 1238: never used?

mjtalbot (approved):
lgtm. i guess a golden test is knocking on our door for stuff like this eh.

js/src/rive.ts
avivian: `alignRenderer()` is looking after this now

PR #146, opened 2021-12-09
Skia Renderer + Assets + rive_light.js

luigi-rosso:
Added some more documentation to the readmes to describe the differences between:

- rive.js
- rive_single.js
- rive_light.js
- rive_canvas.mjs
- rive_canvas_single.mjs
- rive.min.js


luigi-rosso:
Finally ready to review. There are still some updates that need to be done to the non-bundled examples, but those'll be easier once this build goes up to NPM (so they can reference the live build).

avivian:
Major version needs to be bumped!

avivian (approved):
LGTM, will merge to get into react and then website

PR #152, opened 2021-12-13
Remove rogue npm call

avivian:


luigi-rosso:
Oopsy poopsy, thanks @avivian :)

PR #154, opened 2021-12-14
New NPM packages

luigi-rosso:
I started building out the new package structures and the build processes for them. There are now 8 packages in this repo:
- @rive/canvas
- @rive/canvas-single
- @rive/canvas-advanced
- @rive/canvas-advanced-single
- @rive/webgl
- @rive/webgl-single
- @rive/webgl-advanced
- @rive/webgl-advanced-single

Still a few things to do:
- [x] update build scripts to run tests
- [x] update build scripts to bump versions of each of the 8 packages
- [x] update examples


luigi-rosso:
Note that I only updated the parcel example, I still want to update the other "simpler" ones once a version of each package is in NPM.

js/build.sh
```diff
@@ -10,7 +10,7 @@ mkdir -p dist
 # copy the rive wasm js and type definitions files to build dir
 if [ ! -f ../wasm/publish/rive.mjs ]; then 
     pushd ../wasm
-    ./build-all-wasm.sh
+    ./build_all_wasm.sh
```
luigi-rosso, line 13: This is a little pedantic, but in every other repo we take care to use _ instead of - in filenames, so I thought I'd normalize here...

PR #159, opened 2021-12-16
Fix bump version script to use individual repo versions

avivian:


js/bump_version.sh
```diff
@@ -5,5 +5,5 @@
 
 set -e
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-NPM_VERSIONS=`npm show rive-js versions`
```
avivian, line 8: This is the main fix, we need versions for the different repos, not just rive-js

mjtalbot (approved):
lgtm.. i guess

PR #162, opened 2021-12-16
Use FileManagerPlugin to copy declaration files

avivian:
Use FileManagerPlugin as it will wait until the end of the builds before trying to copy the declaration files


luigi-rosso:
Oh interesting, the copy plugin was running before the build ended?! Or is it because you need to wait for all the builds to end?

avivian:
Looks like it was somewhat random, but most of the time copy would run before the build had finished. Didn't catch this locally as build already existed during testing

luigi-rosso:
That's nuts...nice catch.

PR #163, opened 2021-12-20
Update readmes

luigi-rosso:
Noticed most of the README is out of date with the new packages we added last week. I figured this'd be a decent first draft/temporary patch until devrel can help more here.


luigi-rosso:
Adding you here too @neurowave so you can take a peek at the new names and general phrasing of the various options.

PR #167, opened 2022-01-06
Docs: Update the main README documentation for simpler getting started docs and move the separated packages docs into another md file

zplata:
Looking to simplify the README docs for an easier "get started" experience read-through, now that we have different packages for web.

I think the separate runtime packages deserve their own `MD` file to explain further each of their purpose (and when we inevitably add more docs to it). We can link to it from the installation section. This new `WEB_RUNTIMES.md` is mainly just moved over from the main README, with a few tweaks to the wording here and there. I put `@rive-app/webgl` as the current recommendation for now.

Re: the main README:
- Updated the wording on the intro sections and installation and getting started piece
- Added a blurb at the end of the getting started section about forking a simple [CodeSandbox example](https://codesandbox.io/s/rive-plain-js-sandbox-1ddrc) I spun up, so it's easy to plop in the rive file and go.
- Moved the v6 and WASM sections to the bottom
- Removed references to rive-js / rive.js
- Any formatting nits were I think done by my Prettier vscode plugin?


luigi-rosso (approved):
LGTM! Only recommendation I have is to maybe add a visual to call the CodeSandBox demo out even more:
image

Regarding formatter, I think we're open to suggestions. Are you some standard one that you recommend? I don't think any of us made any concerted effort to standardize on markdown formatting, so this is a good opportunity to do so!

zplata:
> LGTM! Only recommendation I have is to maybe add a visual to call the CodeSandBox demo out even more: image

Good call! Added https://github.com/rive-app/rive-wasm/pull/167/commits/3a6e8b7679b588374aa6d3d76677e51f515cb07e to get a photo in.

> Regarding formatter, I think we're open to suggestions. Are you some standard one that you recommend? I don't think any of us made any concerted effort to standardize on markdown formatting, so this is a good opportunity to do so!

I generally have used [Prettier](https://prettier.io/) in the past for opinionated formatting, and think it works quite well, not just for markdown, but JS/TS, JSON, and others. I can set up a prettier config file in a follow-up change!

PR #171, opened 2022-01-12
Bump rive-cpp with skia wasm simd build fix

avivian:
This is to enable the `@rive-app/webgl*` runtimes to work on safari/webkit


umberto-sonnino (approved):
👍 
I see rive-cpp is already merged.

PR #173, opened 2022-01-13
Fallback context to webgl if webgl2 not available

avivian:
We no longer call `createContext`, as it's not doing what we want it to do. While there's some fixes for different browser versions within that function but I don't think there's anything major there we need to work. You can check for yourself here: https://github.com/emscripten-core/emscripten/blob/main/src/library_webgl.js#L575-L709

### DIFF
--- a/wasm/js/skia_renderer.js
+++ b/wasm/js/skia_renderer.js
@@ -14,7 +14,14 @@ Module.onRuntimeInitialized = function () {
             'explicitSwapControl': 0,
             'renderViaOffscreenBackBuffer': 0
         };
-        var handle = GL.createContext(canvas, contextAttributes);
+
+        var ctx = canvas.getContext('webgl2', contextAttributes);
+        // Fallback for browsers that don't support webgl2 (e.g Safari 14)
+        if (!ctx) {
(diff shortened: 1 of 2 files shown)


wasm/js/skia_renderer.js
```diff
@@ -14,7 +14,14 @@ Module.onRuntimeInitialized = function () {
...
             'renderViaOffscreenBackBuffer': 0
         };
-        var handle = GL.createContext(canvas, contextAttributes);
+
+        var ctx = canvas.getContext('webgl2', contextAttributes);
+        // Fallback for browsers that don't support webgl2 (e.g Safari 14)
+        if (!ctx) {
+            ctx = canvas.getContext('webgl', contextAttributes);
+        }
+        var handle = GL.registerContext(ctx, contextAttributes);
```
luigi-rosso, line 23: I didn't know this was a thing, much better!!

luigi-rosso (approved):
LGTM!

wasm/js/skia_renderer.js
rivereed: very clear!

rivereed (approved):

PR #176, opened 2022-01-18
Fix lag when using context2d and add fps/frametime metrics to runtime

avivian:
* Removes an extra `renderer.save()`. This was causing the context stack grow which caused the eventual lag. You can replicate the lag easily by adding a couple of extra calls.
* Add an fps/frametime counter to the high level runtime. We might want to make this optional for performance reasons, however the overhead seems really small.
* Fixed some watch issues with webgl

### DIFF
--- a/js/examples/hello_world/index.html
+++ b/js/examples/hello_world/index.html
@@ -11,7 +11,7 @@
             <canvas id="canvas" width="500" height="500"></canvas>
         </div>
 
-        <script src="/dist/rive.min.js"></script>
+        <script src="https://unpkg.com/@rive-app/webgl@latest"></script>
         <script>
             // autoplays the first animation in the default artboard
             new rive.Rive({
--- a/js/examples/hello_world/index.html
+++ b/js/examples/hello_world/index.html
@@ -11,7 +11,7 @@
             <canvas id="canvas" width="500" height="500"></canvas>
         </div>
 
-        <script src="https://unpkg.com/@rive-app/webgl@latest"></script>
+        <script src="/dist/rive.min.js"></script>
         <script>
             // autoplays the first animation in the default artboard
             new rive.Rive({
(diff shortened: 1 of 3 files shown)


js/src/rive.ts
```diff
@@ -1241,7 +1256,6 @@ export class Rive {
     // Canvas must be wiped to prevent artifacts
     renderer.clear();
     // Now save so that future changes to align can restore
-    renderer.save();
```
avivian, line 1244: The offending line!

js/examples/hello_world/index.html
```diff
@@ -11,7 +11,7 @@
             <canvas id="canvas" width="500" height="500"></canvas>
         </div>
 
-        <script src="/dist/rive.min.js"></script>
+        <script src="https://unpkg.com/@rive-app/webgl@latest"></script>
```
mjtalbot, line 14: intended to stay?

mjtalbot: (or is this just testing against latest?)

js/src/rive.ts
mjtalbot: good riddence

js/src/rive.ts
```diff
@@ -1206,16 +1213,24 @@ export class Rive {
...
-    
 
     // Check for any animations that looped
     this.animator.handleLooping();
 
     // Check for any state machines that had a state change
     this.animator.handleStateChanges();
 
+    // Add duration to create frame to durations array
+    this.frameCount++;
```
mjtalbot, line 1225: assuming we're not worreid about a performance impace of this stuff?

mjtalbot (approved):
lgtm! i think the index.html thing might be a testing leftover

js/examples/hello_world/index.html
avivian: didn't work before I changed, as I think the path has changed. I'll revert this back but it needs to be fixed up at some point

js/src/rive.ts
avivian: Doesn't seem to be much, if any, significant overhead from the profiling I've done

PR #177, opened 2022-01-25
Update rive-wasm to use skia_rive_optimized version of skia.

luigi-rosso:
Uses the latest version of rive-cpp which splits out "skia" and "skia_rive_optimized". This ensures that rive-wasm uses "skia_rive_optimized"


rivereed (approved):
We'll need *some* fontmgr enabled for the non-optimized build, though that can happen in a separate PR if you like.

luigi-rosso:
It's enabled in the rive-cpp PR!

zplata (approved):

PR #178, opened 2022-02-05
fix: updated rive-cpp to include lateset version of skia optimized

mjtalbot:
updates cpp, which will bring it to latest
side effect will be a build with head of rive_optimized_cpp branch, so this sneakily also includes that. 

anything other than pr needed to get the wasm/js side released?


avivian (approved):
That should be it unless any paths changed to the skia build scripts. I think Luigi changed that last PR though

umberto-sonnino (approved):

PR #179, opened 2022-02-06
Feat: Add util function on rive object to set canvas drawing board size based on devicePixelRatio

zplata:
We've received a few comments on why the initial render looks pixelated using the JS runtime. Perhaps instead of relying on consumers to set width/height attributes (usually the reason), we can set them based on the size of the canvas and `window.devicePixelRatio`. This seems to be similar to what the `rive-react` runtime is doing implicitly, but that runtime does a tad more style manipulating because it also controls a container div wrapping the canvas.

Consumers can use this API in a window resizing function to continuously scale the drawing board size of the canvas. If folks initially set a width/height attribute on the canvas, we don't initially run this util (so as not to break existing implementations).

Open to other thoughts here too

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -845,6 +845,11 @@ export interface RiveParameters {
   onLoop?: EventCallback,
   onStateChange?: EventCallback,
   /**
+   * If true, the canvas will scale the drawing board size based on a multiplier of the devicePixelRatio. 
+   * Defaults to true.
+   */
+  useDevicePixelRatio?: boolean,
+  /**
    * @deprecated Use `onLoad()` instead
    */
   onload?: EventCallback,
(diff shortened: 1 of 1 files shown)


luigi-rosso:
This is great, nice documentation. I wonder if we should just default to always using the devicePixelRatio. Just throwing it out there, should we even have it as an option?

zplata:
I'm ok with that - we may want to deprecate the flag in React for the next major if so

avivian (approved):
Looks good to me. We should check how the react runtime works with this and make any modifications required there when we update!

zplata:
@avivian pushed up a small change per @luigi-rosso's suggestion to just use `devicePixelRatio` automatically. We'll probably want to deprecate the option next in react for parity

Issue #183, opened 2022-03-01
Update rive-wasm to latest renderer abstraction from rive-cpp


luigi-rosso:
We made some shiny new changes to the renderer abstraction in rive-cpp, we need to do a little work to make sure both the WebGL and context2d renderers are updated to support them. I've started working on this in the [update_rendering_abstraction](https://github.com/rive-app/rive-wasm/tree/update_rendering_abstraction) branch.

PR #184, opened 2022-03-01
Update rendering abstraction

luigi-rosso:
We recently updated some of the rendering abstractions from the C++ side. This ensures that both WebGL and Context2D renderers are compatible. 

Fixes issue #183


mikerreed (approved):
tested locally

PR #188, opened 2022-03-16
Shared GL Context

luigi-rosso:
You can now use a Shared GL context by supplying an optional second parameter to makeRenderer:
```
const renderer = rive.makeRenderer(canvas, true);
```

This currently hasn't been bubbled up to the high-level (rive.ts) API. Perhaps @avivian or @zplata can help take a stab at that. The idea is to make this functionality opt-in only for now. Calling makeRenderer the old way will supply you with a single GL context as it used to:
```
const renderer = rive.makeRenderer(canvas);
```


zplata:
@luigi-rosso lgtm; I can take on the high-level API to pass in `false` by default. Just seeing a conflicting file here in the submodule

luigi-rosso:
> @luigi-rosso lgtm; I can take on the high-level API to pass in `false` by default. Just seeing a conflicting file here in the submodule

Awesome, thanks @zplata! I updated to the latest submodule and re-ran everything locally. Still looks good. Just going to wait for the tests to complete...

zplata (approved):

wasm/src/bindings.cpp
```diff
@@ -296,11 +296,7 @@ class WebGLSkiaRenderer : public rive::SkiaRenderer {
     m_Canvas = m_Surface->getCanvas();
   }
 
-  void clear() {
```
mjtalbot, line 299: gotta wonder why we did this before, but i guess this is why its good to cover these kind of things with comments / tests. calling clear sounds pretty good without more context on what looks like a workaround

mjtalbot: but i guess this now also calls something slightly different :P

umberto-sonnino (approved):

csmartdalton:
Sweeeeeet!!!

PR #189, opened 2022-03-17
Feat: New useOffscreenRenderer param set in the high-level API constructor set at false by default

zplata:
This adds high-level API support for using the shared WebGL context, introduced in: https://github.com/rive-app/rive-wasm/pull/188

It's set to false by default, so no breaking change here.

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -300,6 +300,21 @@ The following are more properties exposed on the Rive object during instantiatio
 - _isPaused_: are all animations paused?
 - _isStopped_: are all animation stopped?
 
+### Other Notes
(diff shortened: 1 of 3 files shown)


js/README.md
```diff
@@ -300,6 +300,21 @@ The following are more properties exposed on the Rive object during instantiatio
...
+const foo = new Rive({
+    src: "truck.riv",
+    canvas: canvasInstance,
+    animations: "idle",
+    layout: new Layout({fit: "cover", alignment: "center"}),
+    autoplay: true,
+    useOffscreenRenderer: true,
+});
+```
+Note that in a our next major version, this may be turned on by default.
```
mjtalbot, line 316: 🥳

mjtalbot (approved):
lgtm

PR #191, opened 2022-03-23
Initial work to make js/build.sh support incremental builds

csmartdalton:
Builds the various targets in their own directories. Moves the "clean"
step to a block that can be commented out in order to enable incremental
builds. Build steps for individual targets can also be commented out now
that their artifacts persist if we skip the clean step.


luigi-rosso (commented):
Nice! LGTM.

luigi-rosso (approved):

PR #192, opened 2022-03-23
Update build_wasm.sh to use the actual number of CPUs available

csmartdalton:


umberto-sonnino (approved):
We should propagate this to our other build scripts as well! 
I'm guessing you're running on Windows with a UNIX shell, is this compatible across platforms?

PR #194, opened 2022-03-24
Fix alignment issues caused by clearing inside of a save/restore.

luigi-rosso:
Fixes the alignment issue we've been seeing in Rive-React.

Issue was caused by calling renderer.clear() inside of a save()/restore(). The clear op could recreate the surface (which blows aways the canvas context) so then you'll end up with an extra align always on the stack.


mikerreed (approved):
might think about different names in the future
- clear for erase
- resize or something like that

zplata (approved):

PR #196, opened 2022-03-26
Chore: Clean up parcel example test to be more dynamic and add docs on setup

zplata:
Making the parcel example in the `wasm` examples more testing-friendly, by making the num of animations and size of them drawn on the screen determined by query params, and adding a few more examples


csmartdalton:
LGTM!

csmartdalton (approved):

PR #197, opened 2022-03-29
Render offscreen canvases in an atlas

csmartdalton:


wasm/src/bindings.cpp
```diff
@@ -321,13 +323,50 @@ class WebGLSkiaRenderer : public rive::SkiaRenderer {
                kRGBA_8888_SkColorType, nullptr, nullptr)
         .release();
   }
+
+  void clipRect(float l, float t, float r, float b) {
+    auto rect = rive::makeRenderPath();
```
luigi-rosso, line 328: I think we might need to delete this at some point. The ones Rive makes get deleted when the artboard is destroyed.

csmartdalton:
Below is the effect of this change. We may want to selectively disable the atlas on Firefox, but it is otherwise a positive change all around, especially on Safari.

<html>
<body>
<google-sheets-html-origin><style type="text/css"></style>

http://localhost:1234/?numCanvases=40 |   |  
-- | -- | --
  | No Atlas | Atlas
Mac Chrome | 58 | 100
Mac Safari | 10 | 44
Mac Firefox | 25 | 3
Win Chrome | 45 | 60
Win Edge | 41 | 52
Win Firefox | 7 (also looks wrong) | 2 (looks right)


</body>
</html>[http://localhost:1234/?numCanvases=40](http://cmm:1234/?numCanvases=40)		
	No Atlas	Atlas
Mac Chrome	58	100
Mac Safari	10	44
Mac Firefox	25	3
Win Chrome	45	60
Win Edge	41	52
Win Firefox	7 (also looks wrong)	2 (looks right)

wasm/src/bindings.cpp
csmartdalton: Good catch! Done.

wasm/js/skia_renderer.js
```diff
@@ -4,47 +4,54 @@ Module.onRuntimeInitialized = function () {
...
-                glCanvas.width = canvas.width;
-                glCanvas.height = canvas.height;
+        this._drawList = [];
+        this._saveCount = 0;
+
+        this['clear'] = function () {
+            this._drawList = [];  // TODO: what happens on clear if there is a clip???
+            _pendingRenderers.delete(this);
+            if (this._saveCount != 0) {
+                throw "Dunno how to handle a save count in clear()";
```
csmartdalton, line 15: This was more of a question.. Is this scenario possible?

wasm/js/skia_renderer.js
```diff
@@ -4,47 +4,54 @@ Module.onRuntimeInitialized = function () {
...
         this['flush'] = function () {
-            _offscreenGL.flush();
-            const ctx = this._ctx;
-            ctx.globalCompositeOperation = 'copy';
-            ctx.drawImage(_offscreenGL._canvas, 0, 0);
+            while (this._saveCount > 0) {
+                this['restore']();
+            }
+            if (_pendingRenderers.size == 0) {
+                setTimeout(flushPendingRenderers, 0);
```
csmartdalton, line 49: I really dislike setting a timeout here. Could we alternatively (now or later) require the user to call "rive.flushOffscreenCanvas()' during their animation callback, once they're done operating on all their offscreen canvases?

wasm/js/skia_renderer.js
luigi-rosso: Current expectation is clear should always reset/clear the entire frame buffer. Separate from clip, but I think we had one case where we were calling clear after a transform but we fixed that, we should be mindful of not doing that...

wasm/js/skia_renderer.js
csmartdalton: Done.

After cleaning up the requestAnimationFrame stuff, this isn't as good as I had hoped, but still a 2x boost on safari.

csmartdalton:
This is ready to land. Let's get it live!

zplata (approved):

PR #200, opened 2022-03-31
Prefer 8 samples for Skia's internal path atlas

csmartdalton:
We don't seem to always use high resolution canvases yet (via
window.devicePixelRatio). Always request 8 samples for the internal path
atlas.

### DIFF
--- a/wasm/src/bindings.cpp
+++ b/wasm/src/bindings.cpp
@@ -325,6 +325,9 @@ public:
 
 WebGLSkiaRenderer *makeSkiaRenderer(int width, int height) {
   GrContextOptions options;
+  // We don't seem to always use high resolution canvases yet (via window.devicePixelRatio).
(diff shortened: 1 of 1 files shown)


wasm/src/bindings.cpp
```diff
@@ -325,6 +325,9 @@ class WebGLSkiaRenderer : public rive::SkiaRenderer {
 
 WebGLSkiaRenderer *makeSkiaRenderer(int width, int height) {
   GrContextOptions options;
+  // We don't seem to always use high resolution canvases yet (via window.devicePixelRatio).
```
zplata, line 328: We try to detect this, but at the high-level API.
https://github.com/rive-app/rive-wasm/blob/master/js/src/rive.ts#L1435

zplata (approved):

PR #202, opened 2022-03-31
Fix: Make canvas and animation order deterministic and don't use offscreen renderer in FF

zplata:
Fixing a small bug in the parcel example. Also a request to test out no offscreen renderer in firefox as a few of us are able to repro a 300 webgl context limit.


csmartdalton (approved):
Do we want to add a comment about why we don't use the offscreen renderer in Firefox?

PR #203, opened 2022-04-01
Add rive.requestAnimationFrame

csmartdalton:
This will allow us to defer offscreen canvas renderings until they can
all be batched together in a single atlas.

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -1629,7 +1629,11 @@ export class Rive {
    */
   public startRendering() {
     if (this.loaded && !this.frameRequestId) {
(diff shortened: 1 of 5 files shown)


zplata (commented):
This will be a breaking change mainly for the advanced API users, as they'll now have to use `rive.requestAnimationFrame`. The high-level isn't breaking mainly bc we control that internally and end-users don't have to do anything. But because all our npm releases are tied together, they'll all be bumped up a major if we go that route. I'd also suggest we do a `beta` release too. 

Edit: Actually...maybe it's not too bad? This only affects folks who use the advanced API's and have the offscreenRenderer set to `true`, which was _just_ introduced.

PR #210, opened 2022-04-20
Batch webgl meshes in an atlas

csmartdalton:
Rather than our previous approach of:

 * Draw mesh in webgl
 * Blit webgl canvas into canvas2d
 * Draw mesh in webgl
 * Blit webgl canvas into canvas2d

We draw all the meshes as an atlas in one batch:

 * Draw 2 meshes in webgl
 * Blit mesh #1 from webgl canvas into canvas2d
 * Blit mesh #2 webgl canvas into canvas2d

This improves performance by quite a bit in M1 Safari.


zplata (approved):

avivian (approved):
rubber stamp!

PR #212, opened 2022-04-24
Add an fps counter to AnimationCallbackHandler

csmartdalton:
This is a simple timer, fixed to the top right corner, that just
measures the speed of our global rive.requestAnimationFrame() callbacks.
It is enabled by calling rive.enableFPSCounter(true).

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -1654,6 +1654,7 @@ export class Rive {
 
   /**
    * Enables frames-per-second (FPS) reporting for the runtime
+   * If no callback is provided, Rive will append a fixed-position div at the top-right corner of the page with the FPS reading
    * @param fpsCallback - Callback from the runtime during the RAF loop that supplies the FPS value
    */
   public enableFPSCounter(fpsCallback?: FPSCallback) {
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -734,6 +734,11 @@ export interface EventListener {
   callback: EventCallback,
 }
 
+/**
+ * FPS Reporting through callbacks sent to the WASM runtime
+ */
+export type FPSCallback = (fps: number) => void;
+
 // Manages Rive events and listeners
 class EventManager {
 
@@ -1645,6 +1650,21 @@ export class Rive {
         this.frameRequestId = requestAnimationFrame(this.draw.bind(this));
       }
     }
+  }
+
+  /**
+   * Enables frames-per-second (FPS) reporting for the runtime
+   * @param fpsCallback - Callback from the runtime during the RAF loop that supplies the FPS value
+   */
+  public enableFPSCounter(fpsCallback?: FPSCallback) {
+    this.runtime.enableFPSCounter(fpsCallback);
+  }
+
+  /**
+   * Disables frames-per-second (FPS) reporting for the runtime
+   */
+  public disableFPSCounter() {
+    this.runtime.disableFPSCounter();
   }
 
   /**
--- a/wasm/examples/parcel_example/index.js
+++ b/wasm/examples/parcel_example/index.js
@@ -100,9 +100,6 @@ async function renderRiveAnimation({ rive, num, hasRandomSizes }) {
   let lastTime = 0;
   let artboard, stateMachine, animation;
 
-  const times = [];
-  const durations = [];
-
   function draw(time) {
     if (!lastTime) {
       lastTime = time;
@@ -110,15 +107,6 @@ async function renderRiveAnimation({ rive, num, hasRandomSizes }) {
     const elapsedMs = time - lastTime;
     const elapsedSeconds = elapsedMs / 1000;
     lastTime = time;
-
-    const before = performance.now();
-    // Update fps
-    while (times.length > 0 && times[0] <= time - 1000) {
-      times.shift();
-      durations.shift();
-    }
-    times.push(before);
-    document.getElementById("fps-value").innerText = times.length;
 
     renderer.clear();
     if (artboard) {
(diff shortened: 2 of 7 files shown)


wasm/examples/parcel_example/index.js
```diff
@@ -147,14 +135,6 @@ async function renderRiveAnimation({ rive, num, hasRandomSizes }) {
     }
     renderer.flush();
 
-    // Update frame time
-    const after = performance.now();
-    durations.push(after - before);
-    // Use average of all recent durations
-    document.getElementById("framems-value").innerText = (
```
csmartdalton, line 154: Due to the various operations that may or may not be deferred, depending on a variety of reasons, I usually find single frame time measurements to be unhelpful and even misleading. I just took it out of here, but can add it back to the global Rive fps counter if it's something we want.

csmartdalton (commented):
@mikerreed @caudetgit @zplata here is the FPS counter we've been waiting for!

Now the user just needs a way to enable it on the site. Any ideas?

wasm/js/animation_callback_handler.js
```diff
@@ -1,6 +1,41 @@
 // Manages a list of animation callbacks to be called in batch.
 // Override this.onAfterCallbacks to get a call once all animation callbacks have been invoked.
 function AnimationCallbackHandler() {
+    function FpsCounter() {
+        const div = document.createElement('div');
+        div.style.backgroundColor = 'black';
+        div.style.position = 'fixed';
```
csmartdalton, line 7: I'm finding that this div gets drawn mostly on top because we add it to the DOM late, but it can be placed underneath "position: absolute" elements, even if I set style.zOrder to something like 999999. There is probably a way to make sure it's always on top, but I'm starting to step outside my realm of experience.

wasm/js/renderer.js
```diff
@@ -717,5 +717,7 @@ Rive.onRuntimeInitialized = function () {
             _animationCallbackHandler.requestAnimationFrame.bind(_animationCallbackHandler);
     Rive['cancelAnimationFrame'] =
             _animationCallbackHandler.cancelAnimationFrame.bind(_animationCallbackHandler);
+    Rive['enableFPSCounter'] =
```
zplata, line 720: it may be frowned upon...but in this case, I wonder if we might want to put this on the `window` object. I'm thinking in cases where for the rive site for example, we have multiple high-level RIve objects existing at the same time. Because this is more of a global tool - we could set something like `enableRiveFPSCounter()` and then a client with multiple of these Rive objects doesn't need to reason with which one to set the API to `true` here. wdyt? also @avivian

wasm/js/animation_callback_handler.js
```diff
@@ -41,6 +76,18 @@ function AnimationCallbackHandler() {
         });
 
         this.onAfterCallbacks();
+
+        if (_fpsCounter) {
+            _fpsCounter.frameComplete();
+        }
+    }
+
+    this.enableFPSCounter = function(enable) {
```
zplata, line 85: I wonder if we can make this a bit more flexible. what if we had a second parameter here called `options` which is an object. In the object, we could define `elementToDisplayTo` where the consumer can supply a DOM element to push the fps to. Then you don't have to deal with creating the element here and reasoning about styling.

zplata, line 85: Another idea to add in the ^^ `options` is `fpsEventName` where the consumer passes along an event name. Then instead of (or alongside of) setting the innerhtml with the FPS number, we dispatch a custom event with the event name passed in. The consumer would have an event listener that allows them to deal with the fps number however they want. This would be useful for example where instead of setting HTML for this value, they would rather have it console logged, or sent to a logging tool, etc.

wasm/js/renderer.js
csmartdalton: I don't think that would work because this can only measure fps for "rive.requestAnimationFrame" calls on the given rive object.

Would it be possible for us to only have a single high-level rive object? What's the reason for having multiple? Having one is definitely better for offscreen webgl performance.

wasm/js/animation_callback_handler.js
csmartdalton: That sounds nice. We probably want to ultimately provide this to artists so they can see the fps of their animations, so it's worth making it look polished.

I also don't want to worry about making my own FPS element for a simple test page. Can we keep the simple fps counter  approach for when the object or elementToDisplayTo are not provided?

csmartdalton: Let's just do this idea, actually.

* User provides a callback that receives an FPS callback once/second.
* If no callback is provided, we add a simple div to the page and dump FPS.

zplata: Makes sense to me

csmartdalton: Done!

wasm/js/renderer.js
zplata: Every high-level Rive object is tied to a specific rive file. This is how the parcel example works too. So if a user wants to lay out 5 animations from 5 different files, we'll have 5 render loops going on, and 5 fps times to report. We can certainly display fps per rive animation

zplata (approved):
Added the high-level API connection to get the API hooked up - feel free to merge if it looks ok to you

js/src/rive.ts
```diff
@@ -1647,6 +1652,21 @@ export class Rive {
     }
   }
 
+  /**
+   * Enables frames-per-second (FPS) reporting for the runtime
+   * @param fpsCallback - Callback from the runtime during the RAF loop that supplies the FPS value
```
csmartdalton, line 1657: // If null, Rive will append a fixed-position div at the top-right corner of the page with the FPS reading.?

PR #214, opened 2022-04-28
Fix: Don't flush offscreen renderers if there are none

zplata:
For WebGL, if a user does not set `useOffscreenRenderer: true` as part of the attributes in instantiating Rive, they'll get a constant stream of errors due to a null reference in the `flushoffscreenRenderers` which is called in the render loop.

So doing a quick check here in the skia renderer

### DIFF
--- a/wasm/js/skia_renderer.js
+++ b/wasm/js/skia_renderer.js
@@ -127,6 +127,9 @@ Module.onRuntimeInitialized = function () {
 
     // Draws the offscreen renderers all together in a single atlas.
     function flushOffscreenRenderers() {
+        if (!_offscreenGL) {
+            return;
+        }
         const maxRTSize = _offscreenGL._maxRTSize;
 
         // Gather some atlas stats and transfer the pending renders to a sortable array.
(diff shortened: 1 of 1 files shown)


wasm/js/skia_renderer.js
```diff
@@ -261,7 +261,9 @@ Module.onRuntimeInitialized = function () {
             _animationCallbackHandler.cancelAnimationFrame.bind(_animationCallbackHandler);
     Rive['enableFPSCounter'] =
             _animationCallbackHandler.enableFPSCounter.bind(_animationCallbackHandler);
-    _animationCallbackHandler.onAfterCallbacks = flushOffscreenRenderers;
+    if (_offscreenGL) {
```
csmartdalton, line 264: Instead of doing the check here, can we instead do:

```
if (!_offscreenGL) {
    return;
}
```

From flushOffscreenRenderers?

I"m thinking about the case where the user adds a offscreen WebGL after the fact, after this code has executed. In that case I think we wouldn't get the flushOffscreenRenderers hook that we need.

csmartdalton (changes requested):

wasm/js/skia_renderer.js
csmartdalton: Also, thanks for catching this and taking care of it!

zplata: np! Pushed up that change

csmartdalton (approved):

PR #216, opened 2022-04-30
Breaking: DRAFT - bumping cpp

mikerreed:


mikerreed:
wondering if we can/should try rolling part-way forward, so we can merge in few changes per PR (e.g. just take the instances change, or the File/BinaryReader change).

mikerreed:
@zplata

PR #218, opened 2022-05-02
Change Artboard.js to always be an instance

mikerreed:
Just changed the bindings to return the instance, and removed the "instance()" method from js

Lots of other "upgrades" to follow, but taking small steps...


avivian (approved):
LGTM!

PR #223, opened 2022-05-12
Feat: Add interactions functionality

zplata:
Can test this through `js/examples/_frameworks/parcel_example_canvas` or the parcel example in `wasm/examples/parcel_example`. Should be some docs in the readme for running those.

Works for canvas2d and WebGL! Not bumped to the latest and greatest cpp yet due to some more refactor work, but this should get us the minimum for support on interactions.


js/src/rive.ts
```diff
@@ -1452,6 +1456,65 @@ export class Rive {
     }
   }
 
+  public handleTouchEvents() {
```
csmartdalton, line 1459: What if we called this something like "registerTouchEvents"?

"handleTouchEvents" made me wonder if it was intended to be some sort of main loop or something.

wasm/examples/parcel_example/index.js
```diff
@@ -147,8 +216,8 @@ async function main() {
   const numCanvases = parseInt(params.numCanvases || 0) || 19;
   const hasRandomSizes = !!params.hasRandomSizes || false;
   const rive = await RiveCanvas();
-  rive.enableFPSCounter();
-  for (let i = 0; i < numCanvases; i++) {
+  // rive.enableFPSCounter();
```
csmartdalton, line 219: Is this turned off on purpose?

wasm/js/skia_renderer.js
```diff
@@ -39,11 +39,6 @@ Module.onRuntimeInitialized = function () {
             this._drawList.push(_offscreenGL['align'].bind(_offscreenGL, fit, align, from, to));
         };
 
-        this['computeAlignment'] = function (mat2d, fit, align, from, to) {
-            this._drawList.push(_offscreenGL['computeAlignment'].bind(
-                    _offscreenGL, mat2d, fit, align, from, to));
-        };
-
```
csmartdalton, line 46: 🪓🪓🪓🪓🪓

js/src/rive.ts
```diff
@@ -1452,6 +1456,65 @@ export class Rive {
...
+    }
+
+    const activeStateMachines =
+      this.animator.stateMachines.filter((sm: StateMachine) => sm.playing);
+    
+    function mouseCallback(event: MouseEvent) {
+      const boundingRect = (event.target as HTMLCanvasElement).getBoundingClientRect();
+
+      const canvasX = event.clientX - boundingRect.left;
+      const canvasY = event.clientY - boundingRect.top;
```
csmartdalton, line 1471: Do we need to convert to canvas space when high DPI? i.e., multiply by window.devicePixelRatio?

wasm/examples/parcel_example/index.js
```diff
@@ -62,6 +74,63 @@ async function renderRiveAnimation({ rive, num, hasRandomSizes }) {
   // Don't use the offscreen renderer for FF as it should have a context limit of 300
   const renderer = rive.makeRenderer(canvas, true);
 
+  const activeStateMachines = stateMachine ? [stateMachine] : [];
+  function mouseCallback(event) {
```
csmartdalton, line 78: This looks like duplicated code. Is it possible to somehow share with rive.ts?

js/src/rive.ts
zplata: This calculation happens before the register touch event stuff gets called - where we use the window's dpi API to adjust the `canvas.width` and `canvas.height` so it should be taken into account by this point

wasm/examples/parcel_example/index.js
csmartdalton: Cool -- I'll wait for an update!

wasm/examples/parcel_example/index.js
zplata: Yup, moved to a util - in the latest commit!

csmartdalton (approved):

PR #225, opened 2022-05-12
Simplify clipRect lifetime

mikerreed:
Couldn't see why we held onto the paths until flush, so simplifying the code.... thoughts?

### DIFF
--- a/wasm/src/bindings.cpp
+++ b/wasm/src/bindings.cpp
@@ -396,7 +396,6 @@ class WebGLSkiaRenderer : public rive::SkiaRenderer {
 private:
   sk_sp<GrDirectContext> m_Context;
   SkSurface *m_Surface;
-  std::vector<std::unique_ptr<rive::RenderPath>> m_SavedClipRects;
 
 public:
   WebGLSkiaRenderer(sk_sp<GrDirectContext> context, int width, int height)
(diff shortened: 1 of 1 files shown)


csmartdalton:
I believe what's going on here is caution from the fact that at the JS level, we defer a lot of renderer calls until flush time. But in this code, there is clearly not any deferral going on between when we create the clip and when we use it. (clipPath() doesn't go back up into JS where it might get deferred; it stays in WASM.)

Otherwise, is a Renderer allowed to stash away raw pointers until flush time? The SkiaRenderer doesn't do this, but is it generally allowed by rive::Renderer? (If so I would definitely advocate for a ref counted model...)

mikerreed:
> I believe what's going on here is caution from the fact that at the JS level, we defer a lot of renderer calls until flush time. But in this code, there is clearly not any deferral going on between when we create the clip and when we use it. (clipPath() doesn't go back up into JS where it might get deferred; it stays in WASM.)
> 
> Otherwise, is a Renderer allowed to stash away raw pointers until flush time? The SkiaRenderer doesn't do this, but is it generally allowed by rive::Renderer? (If so I would definitely advocate for a ref counted model...)

The renderer can *not* hold on to any raw pointers.

mikerreed:
> 

My read is that clipRect is calling through to SkiaRenderer's clipPath...

void SkiaRenderer::clipPath(RenderPath* path) {
    m_Canvas->clipPath(reinterpret_cast<SkiaRenderPath*>(path)->path(), true);
}

It is passing though SkPath to canvas->clipPath() which copies the skia path, so I think we're good to delete the RenderPath in its stack-frame.

PR #227, opened 2022-05-13
Roll cpp forward (including Factory)

mikerreed:
This is based on https://github.com/rive-app/rive-wasm/pull/226

- Pass factories to File::import()
- Minor changes to some factory signatures
- Minor API change to matrix.invert()


zplata (approved):
LGTM! Thanks for taking this on!

PR #229, opened 2022-05-16
Maint: Bump rive-cpp to incorporate name length fix

zplata:
This brings us to the latest head of cpp


zplata:
Good call @mjtalbot - added!

PR #233, opened 2022-05-23
Docs: Update individual npm README docs

zplata:
Updating these README files that display on the npm package homepages for each of these dependencies we output


js/npm/webgl_single/README.md
```diff
@@ -1,16 +1,19 @@
...
 ```
 npm install @rive-app/webgl-single
 ```
-An easy to use high level Rive API using the WebGL renderer. This lets Rive squeeze every last ounce of performance from the hardware and provide some newer advanced rendering features which will not be available to the Canvas renderers. 
+An easy-to-use high-level Rive API using the [WebGL](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API) renderer. This runtime will support Rive's upcoming advanced rendering features which may not be available to the Canvas renderers. Some benefits of this package:
 - Highest fidelity with edit-time experience.
-- Best performance across all devices.
-- Support for upcoming features like mesh deformations. 
+- Support for future advanced rendering features
+- Web Assembly (WASM) is part of the JS bundle; no need to make a request to load it in at runtime
```
csmartdalton, line 13: I'm not an editor, but I thought the rule for semicolons was that both clauses need to stand as complete sentence on their own. Is that also the case if you're in a bulleted list? Do we have an editor to ask? ¯\_(ツ)_/¯

"Web Assembly (WASM) is part of the JS bundle; there is no need to make a request to load it in at runtime"

csmartdalton (approved):
Edits lgtm.

I have one rambling thought that I don't actually know the answer to, so feel free to land as-is!

PR #240, opened 2022-06-22
Work around mutating fill rules in canvas2d runtime

csmartdalton:
The fill rule (and only the fill rule) on a path object can mutate
before flush(). To work around this, we capture the fill rule at draw
time. It's a little awkward of a solution, so we probably want to rework
this.

https://github.com/rive-app/rive/issues/3816


luigi-rosso (approved):
LGTM, thanks for the quick fix! Let's definitely keep thinking about the long term solution...

PR #242, opened 2022-06-24
Patch: Allow specified animations on instantiation to advance once on autoplay to show a more accurate frame

zplata:
Might not be an absolute solution for now.. as this doesn't really take well when `.stop()` or `.reset()` is called, just due to the way the animations tracked are setup at the moment (`stop()` removes all animations from the list of active animations). But this should help align `autoplay: false` with the first frame of the animation. Totally open to other suggestions here!

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -458,7 +458,7 @@ class Animator {
        } else {
           // Try to create a new animation instance
           const anim = this.artboard.animationByName(animatables[i]);
-          if(anim) {
+          if (anim) {
             this.animations.push(new Animation(anim, this.artboard, this.runtime, playing));
           } else {
             // Try to create a new state machine instance
@@ -503,7 +503,7 @@ class Animator {
    * @param animatables names of the animations and state machines to pause
    * @returns a list of names of the animations and state machines paused
    */
-     public pause(animatables: string[]): string[] {
+    public pause(animatables: string[]): string[] {
       return this.add(animatables, false);
     }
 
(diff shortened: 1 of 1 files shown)


js/src/rive.ts
```diff
@@ -1209,13 +1209,16 @@ export class Rive {
       onSecond?.();
     }
 
+    const isFirstFrame = time - this.lastRenderTime === 0;
```
mikerreed, line 1212: same as firstFrame = (time == this.lastRenderTime);  ?

mikerreed (commented):
Not sure I understand the current bug.

Do we call advance(0) on animations/machines when we first create them? Would that help?

zplata:
> Not sure I understand the current bug.
> 
> Do we call advance(0) on animations/machines when we first create them? Would that help?

@mikerreed  Ah sorry should have elaborated that on the description. Basically, if you instantiate an animation or state machine, but set `autoplay: false` when creating the Rive instance like:
```
new Rive({
  src: "some-file.riv",
  animations: "foo"
  autoplay: false,
})
```

It doesn't pause the animation instance at the first frame as one would expect. That's why in the community page of rive.app, you might see a strange first frame (i.e the thing that shows from the design mode tab) and then when hovering (which calls `.play()` on the animation), it correctly switches to the starting frame of the animation and starts the render loop. The weird glitch is what we want to try and avoid here. I think your solution for advancing and applying on animation instantiation though works better than what I have here, so I'll make that change!

js/src/rive.ts
```diff
@@ -366,6 +377,14 @@ class StateMachine {
     return names;
   }
 
+  /**
+   * Advances the state machine instance by a given time.
+   * @param time - the time to advance the animation by in seconds
+   */
+  public advance(time: number) {
+    this.instance.advance(time);
```
luigi-rosso, line 385: FYI StateMachines will always apply while they advance (advance internally eventually calls apply, potentially multiple times). This is because they have to apply in order for to allow them to set values to use for the next iteration (until they're stable). Just in case you're expecting it not to...

luigi-rosso (commented):
LGTM, just one consideration but I think it's fine as it's what it was doing before too.

luigi-rosso (approved):

js/src/rive.ts
zplata: Ah cool cool, yeah I mainly added this because I noticed in the render loop we have some places where we do `animation.instance.blablabla` and same with state machine, so just abstracting the `.instance` part behind their respective classes so they all just do `animation.advance` or `stateMachine.advance`

PR #243, opened 2022-06-24
Wait for all images to decode before completing load.

luigi-rosso:
I changed load to be async. This is a little funky as I basically replace the native bound "load" with a js one (we were doing this elsewhere for other functions) that returns a Promise.

With the Skia renderer this just returns an already resolved Promise as everything happens synchronously. 

With the Context2D/Canvas renderer we build a "loadContext" that synchronously gets accessed and stored in the load closures for each image. It's hacky but works with multiple loads. I did it this way to not patch anything at the C++ layer.

No breaking changes for the high level API. Small breaking change for the advanced API: they'll now need to await/.then the Promise returned from load.

P.S. formatting is messy, I'll separately PR with a formatter for JS with a note about it in the README.

### DIFF
--- a/js/src/rive_advanced.mjs.d.ts
+++ b/js/src/rive_advanced.mjs.d.ts
@@ -20,7 +20,7 @@ interface RiveOptions {
     StrokeCap: typeof StrokeCap;
     StrokeJoin: typeof StrokeJoin;
   
-    load(buffer: Uint8Array): File;
+    load(buffer: Uint8Array): Promise<File>;
     makeRenderer(canvas: HTMLCanvasElement | OffscreenCanvas, useOffscreenRenderer: boolean) : CanvasRenderer;
     computeAlignment(fit: Fit, alignment: Alignment, frame: AABB, content: AABB): Mat2D;
     mapXY(matrix: Mat2D, canvasPoints: Vec2D): Vec2D;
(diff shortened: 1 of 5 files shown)


js/src/rive_advanced.mjs.d.ts
```diff
@@ -20,7 +20,7 @@ interface RiveOptions {
     StrokeCap: typeof StrokeCap;
     StrokeJoin: typeof StrokeJoin;
   
-    load(buffer: Uint8Array): File;
+    load(buffer: Uint8Array): Promise<File>;
```
luigi-rosso, line 23: Looks like the TS code was already "await"ing load, I just changed it to be a Promise.

wasm/examples/parcel_example/index.js
```diff
@@ -50,7 +50,7 @@ async function renderRiveAnimation({ rive, num, hasRandomSizes }) {
     const bytes = await (
       await fetch(new Request(riveEx.riveFile))
     ).arrayBuffer();
-    const file = rive.load(new Uint8Array(bytes));
+    const file = await rive.load(new Uint8Array(bytes));
```
luigi-rosso, line 53: Anyone using the advanced API will now need to do this (or .then(...)).

zplata (approved):

PR #244, opened 2022-06-24
Prettier Formatter

luigi-rosso:
Depends on https://github.com/rive-app/rive-wasm/pull/243

Standardizes our JS formatting using Prettier. Note that I included a .prettierrc config file as we need to ensure that object key definitions are kept in quotes so that closure doesn't mangle the key names.


zplata (approved):

wasm/README.md
```diff
@@ -10,6 +10,10 @@ Rive's low level web runtime. Use this over the high level JS runtime if you nee
 
 `build_all_wasm.sh` will build all the various permutations of the Rive WASM modules necessary to build rive-js.
 
+# Code Formatting
+
+The C++ code is formatted using ClangFormat. The JS code is formatted using Prettier. Configuration files are included for both. VSCode will pick them up directly if you have the Prettier and ClangFormat extensions installed.
```
csmartdalton, line 15: > Configuration files are included for both

Is this true? I don't see a .clang-format in rive-wasm

PR #245, opened 2022-06-28
Patch: Bump cpp to fix nested artboard opacity issue

zplata:
Bump from cpp to fix the issue where nested artboards by themselves were not rendering


luigi-rosso (approved):
LGTM!

PR #249, opened 2022-07-05
Fix: Add artboard base to bindings to get access to higher-level Artboard methods not available at instance class

zplata:
The `rive::ArtboardInstance` does not inherit some of the methods from `Artboard`, such as `.bone()` or `.node()`, which used to exist before we moved to everything instanced. This is a first attempt solution to try and get those methods to propagate, but there might be a better solution thru emscripten that I'm missing for getting `ArtboardInstance` to inherit properly.

Also adding @luigi-rosso's centaur game example in the `/wasm` examples with the updated low-level API's.


wasm/js/renderer.js
```diff
@@ -642,6 +642,9 @@ Rive.onRuntimeInitialized = function () {
         )
       );
     },
+    "rotate": function (angle) {
+      this._drawList.push(this._ctx["rotate"].bind(this._ctx, angle));
```
csmartdalton, line 646: This needs to update the shadow _matrixStack. Can it call transform instead?

```
const sin = Math.sin(angle);
const cos = Math.cos(radians);
this.transform({"xx": cos, "xy": sin, "yx": -sin, "yy":cos, "tx":0, "ty":0);
```

csmartdalton (changes requested):
lgtm after one request. I'm excited to try out the game!

wasm/js/renderer.js
zplata: @csmartdalton Yeah for sure! changed

zplata: You can pull down this branch and go to `wasm/examples/centaur_game`, run `npm i`, then `npm start`. Same URL as most of the examples, you can go to `localhost:1234` and play it.

csmartdalton (approved):

wasm/examples/centaur_game/index.ts
```diff
@@ -0,0 +1,354 @@
...
+
+      const appleInstance = {
+        x: -appleBounds.maxX + Math.random() * appleBounds.maxX * 3,
+        y: -appleBounds.maxY + Math.random() * -appleBounds.maxY,
+        artboard: aplInstance,
+        machine: appleMachine,
+        explodeTrigger: explodeTrigger,
+      };
+      appleMachine.advance(0);
+      appleInstance.artboard.advance(0);
```
mikerreed, line 69: Can you not call advanceAndApply (just on the machine)?

mikerreed (changes requested):
please hold off landing until I can read more :)

wasm/src/bindings.cpp
```diff
@@ -200,7 +200,8 @@ EMSCRIPTEN_BINDINGS(RiveWASM) {
           allow_raw_pointers())
       .function("artboardCount", &rive::File::artboardCount);
 
-  class_<rive::ArtboardInstance>("Artboard")
+  class_<rive::Artboard>("ArtboardBase");
```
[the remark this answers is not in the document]
zplata: Yes, we can call it ArtboardRaw or something (it doesn't matter) - but seems embind does not like when it's not declared. 
Doc snippet: https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#base-classes

wasm/examples/centaur_game/index.ts
zplata: Oh - this must be new ish? It's not defined in bindings, but looking at the cpp def, yeah, we can.

mikerreed (approved):

Issue #247, opened 2022-06-28
add types in the npm package output


GrandSchtroumpf:
Currently `@rive-app/canvas-advanced` & `@rive-app/webgl-advanced` do not provide type definition.
Would be nice to include these [types](https://github.com/rive-app/rive-wasm/blob/master/js/src/rive_advanced.mjs.d.ts) into the output folder & update the [package.json](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package).

Note: the OffscreenCanvas is not supported by Typescript by default anymore (see [issue](https://github.com/microsoft/TypeScript/issues/45745)). You can either remove it or [add a polyfill](https://www.npmjs.com/package/@types/offscreencanvas) to avoid user compiler from breaking.

Thanks.

zplata:
@GrandSchtroumpf We just merged support for types! I'd try with `@rive-app/canvas-advanced` over WebGL for best performance right now and support. And appreciate the tip about `OffscreenCanvas`, we included a types for that in this latest release `1.0.70`!

PR #250, opened 2022-07-11
Maint: Add types to advanced and advanced single canvas packages

zplata:
Adding types for the canvas advanced packages. WebGL should follow, but we need to add support for `translate` and `rotate` on that renderer. Enabling this for now. Also fixing up some stale types that were incorrect.

Should address #247


js/src/rive_advanced.mjs.d.ts
```diff
@@ -42,6 +45,8 @@ interface RiveOptions {
     clipPath(path: RenderPath): void;
     clear(): void;
     flush(): void;
+    translate(x: number, y: number): void;
+    rotate(angle: number): void;
```
csmartdalton, line 49: Curious how translate/rotate methods are related to adding types to the canvas packages. (What does it mean to add types to them?)

zplata: For projects who consume this runtime, if they're using TypeScript, it's an affordance for them when their project compiles (also is good for documentation purposes). One nice affordance as well is for folks using VSCode, they can have nice things like autocomplete bc it can sense the types of certain objects. so When they do:
```
const renderer = rive.makeRenderer(canvas);
...
renderer.rotate(Math.atan2(x, y));
```
The editor doesn't complain because we've specified `rive.makeRenderer()` returns type `CanvasRenderer` which with these changes, would know `.rotate()` is now a function that can be invoked on `renderer`.

We need to still define translate and rotate on the `skia_renderer.js` in a separate change though, so that's why I mention right now, this will satisfy the `canvas` packages.

csmartdalton: Got it. Thanks for explaning that!

csmartdalton (approved):

Issue #251, opened 2022-07-14
OffscreenCanvas fails with "rive.requestAnimationFrame"


GrandSchtroumpf:
When working with the [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) inside a WebWorker, the method `rive.requestAnimationFrame` throw `window is not defined` (which make sense because there is no `window` object inside a WebWorker).
Using the native `requestAnimationFrame` seems to work though.

zplata:
@GrandSchtroumpf We recently fixed this in 1.0.77 - mind trying it to see if your scenario is addressed by this?

PR #254, opened 2022-07-22
fix: use native requestAnimationFrame instead of window to keep compatibility with offscreencanvas usage

zplata:
Addresses #251 so that we're not restrict to window's rAF functions, and can be used on web workers/offscreen canvases.


csmartdalton (approved):
Nice

PR #256, opened 2022-07-29
Fix: Cleanup render loop and instances on Rive cleanup() and set artboard to null

zplata:
Want to set artboard to null after cleanup so any lingering public methods that try and read from artboard don't try and access artboard pointer methods. Also adding `stop()` to cleanup, to properly clean up animation/state machine instances. That way, folks don't have to call `rive.stop()` and `rive.cleanup()` for a true cleanup.

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -1454,8 +1454,16 @@ export class Rive {
     if (this.eventCleanup !== null) {
       this.eventCleanup();
     }
-    this.artboard.delete();
-    // TODO: delete animation and state machine instances
+    // Stop rAF loop if still continuing, and delete all animation and state machine
+    // instances
+    if (this.frameRequestId) {
+      this.stopRendering();
+      this.stop();
+    }
+    if (this.artboard) {
+      this.artboard.delete();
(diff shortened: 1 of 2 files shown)


js/src/rive.ts
```diff
@@ -1454,8 +1454,12 @@ export class Rive {
     if (this.eventCleanup !== null) {
       this.eventCleanup();
     }
-    this.artboard.delete();
-    // TODO: delete animation and state machine instances
+    // Delete all animation and state machine instances
+    this.stop();
+    if (this.artboard) {
```
luigi-rosso, line 1459: Is this just for sanity or was there a case where we were calling cleanup multiple times too?

luigi-rosso (approved):
LGTM, just a question around when we call cleanup.

js/src/rive.ts
zplata: sanity - in case multiple cleanups are called by consumers. I'm seeing some funny business around how React calls cleanup on useEffects, which is where we call `rive.cleanup()`. But also seems a safeguard change to add in case others do this too.

PR #257, opened 2022-08-02
Revert back to skia_rive_optimized for wasm

csmartdalton:


luigi-rosso:
Looks like the versions aren't compatible, we must've updated some includes and API signatures to work with the newest Skia (see failed test).

csmartdalton:
> Looks like the versions aren't compatible, we must've updated some includes and API signatures to work with the newest Skia (see failed test).

I hope it's just that we need https://github.com/rive-app/rive/pull/4100 first

csmartdalton:
This also needs an update to the rive-cpp submodule. I'll wait for https://github.com/rive-app/rive/pull/4100 in order to make that final change

csmartdalton:
Looks like tests are passing now that I've updated the rive-cpp submodule

zplata (approved):

PR #258, opened 2022-08-24
docs: add types to the relevant classes and methods used for the low-level API and render loop usage

zplata:
Adding JSDocs for most of the relevant low-level API classes/methods used for integrating Rive. This should help document usage here, as well as for others using the low-level APIs with Typescript projects/VScode. 

Need to run thru a linter with the new eslint config in a separate change

### DIFF
--- a/js/src/rive_advanced.mjs.d.ts
+++ b/js/src/rive_advanced.mjs.d.ts
@@ -1,10 +1,15 @@
 interface RiveOptions {
-    locateFile(file: string): string
+    locateFile(file: string): string;
   }
-  
+
   declare function Rive(options?: RiveOptions): Promise<RiveCanvas>;
   export default Rive;
-  
+
+  /**
+   * RiveCanvas is the main export object that contains references to different Rive classes to help
+   * build the Rive render loop for low-level API usage. In addition, this contains multiple methods that help
+   * aid in setup, such as loading in a Rive file, creating the renderer, and starting/finishing the render loop (requestAnimationFrame)
+   */
   export interface RiveCanvas {
     Alignment: AlignmentFactory;
     CanvasRenderer: typeof CanvasRenderer;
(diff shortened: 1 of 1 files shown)


js/src/rive_advanced.mjs.d.ts
```diff
@@ -1,10 +1,15 @@
...
-  
+
   declare function Rive(options?: RiveOptions): Promise<RiveCanvas>;
   export default Rive;
-  
+
+  /**
+   * RiveCanvas is the main export object that contains references to different Rive classes to help
+   * build the Rive render loop for low-level API usage. In addition, this contains multiple methods that help
+   * aid in setup, such as loading in a Rive file, creating the renderer, and starting/finishing the render loop (requestAnimationFrame)
```
csmartdalton, line 11: Line length? What column limit are we using here?

js/src/rive_advanced.mjs.d.ts
```diff
@@ -15,50 +20,121 @@ interface RiveOptions {
...
+     */
     load(buffer: Uint8Array): Promise<File>;
-    makeRenderer(canvas: HTMLCanvasElement | OffscreenCanvas, useOffscreenRenderer?: boolean) : CanvasRenderer;
-    computeAlignment(fit: Fit, alignment: Alignment, frame: AABB, content: AABB): Mat2D;
+
+    /**
+     * Creates the renderer to draw the Rive on the provided canvas element
+     *
+     * @param canvas - Canvas to draw the Rive on
+     * @param useOffscreenRenderer - Option for those using the WebGL-variant of the Rive JS library. This uses an offscreen renderer to draw on the canvas, allowing for multiple Rives/canvases on a given screen. It is highly recommended to set this to `true` when using with the `@rive-app/webgl` package
```
csmartdalton, line 43: Should we also wrap this line to 100 characters?

js/src/rive_advanced.mjs.d.ts
```diff
@@ -15,50 +20,121 @@ interface RiveOptions {
...
+     */
+    computeAlignment(
+      fit: Fit,
+      alignment: Alignment,
+      frame: AABB,
+      content: AABB
+    ): Mat2D;
     mapXY(matrix: Mat2D, canvasPoints: Vec2D): Vec2D;
+    /**
+     * A Rive-specific requestAnimationFrame function; this should be used over the global requestAnimationFrame function.
```
csmartdalton, line 67: I might say "instead of". "over" might imply that this is just a suggestion

js/src/rive_advanced.mjs.d.ts
```diff
@@ -15,50 +20,121 @@ interface RiveOptions {
...
+     * @param cb - Callback function to call with an elapsed timestamp
+     * @returns number - An ID of the requestAnimationFrame request
+     */
     requestAnimationFrame(cb: (timestamp: DOMHighResTimeStamp) => void): number;
+    /**
+     * A Rive-specific cancelAnimationFrame function; this should be used over the global cancelAnimationFrame function.
```
csmartdalton, line 73: s/over/instead of/

?

js/src/rive_advanced.mjs.d.ts
```diff
@@ -15,50 +20,121 @@ interface RiveOptions {
...
+     * @param requestID - ID of the requestAnimationFrame request to cancel
+     */
     cancelAnimationFrame(requestID: number): void;
+    /**
+     * Debugging tool to showcase the FPS in the corner of the screen
+     */
```
csmartdalton, line 79: If you don't provide a callback does this method still add a DIV for you to display it in the corner?

js/src/rive_advanced.mjs.d.ts
```diff
@@ -168,68 +372,140 @@ interface RiveOptions {
...
     constructor(stateMachine: StateMachineInstance, artboard: Artboard);
     get name(): string;
+    /**
+     * Returns the number of inputs associated with this state machine
+     * @returns Number of inputs
+     */
     inputCount(): number;
-    input(i: number): SMIInput; 
+    /**
+     * Returns the SMIInput at the given index
```
csmartdalton, line 424: I would call it a "State machine input" in the comment, since the word SMIInput already exists in the code, and doesn't necessarily make it clear what it refers to if you're new to Rive

js/src/rive_advanced.mjs.d.ts
```diff
@@ -168,68 +372,140 @@ interface RiveOptions {
...
+    /**
+     * Returns the name of the state/animation transitioned to, given the index in the array of state changes in total
+     * @param i
+     * @returns Name of the state/animation transitioned to
+     */
     stateChangedNameByIndex(i: number): string;
 
-    // Call with coordinates in Artboard space
+    /**
+     * Registers a coordinate where the pointer down listener watches for in the Artboard space. Internally, Rive may advance a state machine if the listener coordinate is of interest at a given moment.
```
csmartdalton, line 447: Do we need to talk about state machine listeners here? (Are these things the user can register on?)

Or could we just say something as simple as "Notifies the state machine that the pointer has been pressed"?

js/src/rive_advanced.mjs.d.ts
```diff
@@ -300,26 +576,48 @@ interface RiveOptions {
...
   export interface AABB {
     minX: number;
     minY: number;
     maxX: number;
     maxY: number;
   }
-  
+
+  /**
+   * Matrix described by the following:
```
csmartdalton, line 591: Can we call it a "Column-major matrix described by the following"? Or is order not really even a thing here?

Also, I'm going to get super nit-picky and ask what you think of ASCII art like this:

```
| xx  yx  tx |
| xy  yy  ty |
|  0   0   1 |

File: js/src/rive_advanced.mjs.d.ts [hunk 9]
```diff
@@ -71,95 +147,223 @@ interface RiveOptions {
...
+     * @returns A new StateMachineInstance object
+     */
     stateMachineByIndex(index: number): StateMachineInstance;
+    /**
+     * Returns the number of state machines in the artboard
+     * @returns Number of state machines on the Artboard
+     */
     stateMachineCount(): number;
+    /**
+     * Returns a reference for a Bone object of a given name
```
csmartdalton, line 291: If I don't know how a bone works, is there somewhere I can jump to read up on it?

csmartdalton (commented):
LGTM with a few nits

Wow, this really transformed the code into library that feels polished and high quality!

My major concern is line wrapping. Do you think we should manually wrap all the comments at 100 columns?

js/src/rive_advanced.mjs.d.ts
csmartdalton: Or even, "this must be used instead of"

js/src/rive_advanced.mjs.d.ts
zplata: Prettier defaults to 80, I totally missed this didn't apply to comments on save.

But on further look, seems like they [don't enforce](https://prettier.io/docs/en/rationale.html#print-width) on comments. So we can configure eslint to error out if past like 100.

js/src/rive_advanced.mjs.d.ts
zplata: ah right, good call!

js/src/rive_advanced.mjs.d.ts
zplata: Yeah I like that verbiage - this one was weird to try and document

js/src/rive_advanced.mjs.d.ts
zplata: Can add that!

js/src/rive_advanced.mjs.d.ts
zplata: Honestly not much functionality here that we expose, beyond coordinates and a length. But we can maybe link to the Bone feature in the editor (our help docs center)

js/src/rive_advanced.mjs.d.ts
zplata: oh right! totally forgot thats a thing. this type needs to be altered to show the callback - but yeah, it still does that

zplata:
> LGTM with a few nits
> 
> Wow, this really transformed the code into library that feels polished and high quality!
> 
> My major concern is line wrapping. Do you think we should manually wrap all the comments at 100 columns?

100-len column change: https://github.com/rive-app/rive-wasm/pull/258/commits/bb8fd735621bdc0a0eeaaa437a110a83dfe8b977
Wording changes: https://github.com/rive-app/rive-wasm/pull/258/commits/79c50f6067bdb119685f454b2fe51b4af9e16a93

WIll need to add ESLint config to enforce the max-len of 100 for a column, but might do it in a separate PR since it will most likely affect the other TS files as well.

js/src/rive_advanced.mjs.d.ts
csmartdalton: "passes the fps count to the callback instead"?

Or does it still make the new div if you give a callback?

js/src/rive_advanced.mjs.d.ts
```diff
@@ -1,325 +1,669 @@
...
+   * Notifies the state machine that the pointer has pressed down at the given coordinate in
+   * Artboard space. Internally, Rive may advance a state machine if the listener coordinate is of
+   * interest at a given moment.
+   * 
+   * @param x - X coordinate
+   * @param y - Y coordinate
+   */
+  pointerMove(x: number, y: number): void;
+  /**
+   * Registers a coordinate where the pointer up listener watches for in the Artboard space.
```
csmartdalton, line 500: Notifies the state machine that the pointer has released at ...

csmartdalton (changes requested):
One more tweak

js/src/rive_advanced.mjs.d.ts
```diff
@@ -1,325 +1,669 @@
...
+   * Notifies the state machine that the pointer has pressed down at the given coordinate in
+   * Artboard space. Internally, Rive may advance a state machine if the listener coordinate is of
+   * interest at a given moment.
+   * 
+   * @param x - X coordinate
+   * @param y - Y coordinate
+   */
+  pointerDown(x: number, y: number): void;
+  /**
+   * Notifies the state machine that the pointer has pressed down at the given coordinate in
```
csmartdalton, line 491: that the pointer has **moved to** the given coordinate

csmartdalton (changes requested):

js/src/rive_advanced.mjs.d.ts
zplata: Nope, it doesn't! I'll clarify that

js/src/rive_advanced.mjs.d.ts
zplata: bleh - good catch on these!

csmartdalton (approved):

PR #259, opened 2022-08-30
Remove node from emscripten target environments

avivian:
By default, emscripten targets web, webview, worker and node (https://github.com/emscripten-core/emscripten/blob/6d3e1a9be15bc9ffbeb7f6ec81942bfef96336ad/src/settings.js#L615-L640). By targeting node, emscripten tries to pull in the process, fs and path dependencies if running on a node environment. jspm, which is the package CDN used by framer, automatically adds browser polyfills for these, which fail when framer tries to statically generate sites with rive files.

This config changes remove the node target, which fixes the above problem.


mjtalbot (approved):
lgtm!

Issue #131, opened 2021-08-11
Does not work with strict Content Security Policy (CSP) that blocks unsafe-eval (new Function)


localpcguy:
### Description
An error is thrown if used in an environment where a CSP does not allow `'unsafe-eval'` due to the use of `new Function` in the `createNamedFunction` method currently [at line 693 of the published rive.lean.js file](https://github.com/rive-app/rive-wasm/blob/1007e765f4fb5cbe8dde8d408edcd01764a0d379/wasm/publish/rive.lean.mjs#L693).

```
function createNamedFunction(name, body) {
  name = makeLegalFunctionName(name);
  return new Function("body", "return function " + name + "() {\n" + '    "use strict";' + "    return body.apply(this, arguments);\n" + "};\n")(body);
}
```

The error looks like this (will be slightly different depending on your CSP and the browser viewed in:
```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' ... ;"
```

### Provide a Repro
This should be fairly simple to repro by adding the following CSP in the file that the Rive script is being used in (assuming the Rive script is self hosted, but update localhost to be the domain where the rive script is loaded from to allow it if not.)
```
<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'self' localhost; connect-src: data:;"
```

#### Source `.riv`/`.rev` file
Individual RIV file is not relevant to this issue

### Expected behavior
Expected that the Rive scripts do not try to use `eval` or `new Function` and work as expected without errors when used in an environment with a strict CSP that blocks the use of `'unsafe-eval'`.

### Browser & Versions (please complete the following information)
Any browser that parses and reads Content Security Policies (any modern browser including Chromium based browsers, Firefox and Safari).

darkshredder:
I'm facing the same issue. Any updates on this?

localpcguy:
I have not seen an update to date. I'd be interested in at least taking a look or even stab at fixing it if someone would be able to point me to the actual source files that are compiled/transpiled into the code shown above.

zplata:
This is coming from generated code from the [enscriptem](https://github.com/emscripten-core/emscripten/issues/12910) library that compiles our underlying web assembly and JS-binding "glue" code.

[...]

PR #263, opened 2022-09-16
patch: Add dynamic execution flag to emcc build to allow for wasm-unsafe-eval in csp

zplata:
Should help address https://github.com/rive-app/rive-wasm/issues/131

If folks set CSP policies that block `unsafe-eval` scripts (i.e use of `new Function()` or `eval()`), they may have issues rendering Rives because our WASM (built using Emscripten) that has binding code to JS includes some `new Function()` code as part of Emscripten's inner-workings around binding. There's [some effort](https://github.com/emscripten-core/emscripten/pull/17296) on Emscripten's side to remove some of these pieces, but the guidance for consumers as seen in [this issue](https://github.com/WebAssembly/content-security-policy/issues/7) is to allow `wasm-unsafe-eval` in the CSP. This alone however still doesn't solve everything. We need to set this `DYNAMIC_EXECUTION=0` flag to prevent the use of `new Function()` or `eval()` in Emscripten's native binding code during build. The pairing of this fix in our WASM build setup, and the consumer setting `wasm-unsafe-eval` should get Rives running in web apps if it were blocked before. While not perfect, it's better than setting `unsafe-eval` for sure as a content policy. 

If this takes, we'll document this in our JS runtimes gitbook section

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup emsdk
         uses: mymindstorm/setup-emsdk@v11
         with:
-          version: 3.0.0
+          version: 3.1.20
           actions-cache-folder: "emsdk-cache"
       - name: Install premake
         run: |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ git checkout origin/master
 cd ../../..
 ```
 
-4. Install [Emscripten](https://emscripten.org/docs/getting_started/downloads.html). We build against 3.0.0 in rive-wasm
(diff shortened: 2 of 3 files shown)


csmartdalton:
Could you add a note about what is causing this error in Rive, and what it means for our code to enable `wasm-unsafe-eval`? I read that and it sounds scary to enable 😱

zplata:
@csmartdalton yeah good point! Sorry, didn't really document the issue clearly. I updated the description.

We're not enabling any `wasm-unsafe-eval` on our side; this is a CSP policy attribute set from the user side of things on their web apps if they decide to set content security policies so that their apps aren't blocking our WASM from running

csmartdalton (approved):
Thanks for clarifying! I understand what's going on now

Issue #265, opened 2022-09-19
rive advanced canvas has missing and incorrect types for Vec2D


jrabek:
In order to use the code from rive-canvas that shows how to convert touch points into the coordinate system of the artboard, it is necessary to construct a 2d vector and also get the values from it.

Currently `new rive.Vec2D` in Typescript isn't possible since the type lacks a constructor that accepts x and y.

Also the return types of x and y are void.  Also seems strange that they are functions and not just properties like the other math types in Rive.

https://github.com/rive-app/rive-wasm/blob/master/js/src/rive_advanced.mjs.d.ts#L660
```
/**
 * Rive Vector class
 */
export declare class Vec2D {
  /**
   * Returns the x coordinate of the vector
   */
  x(): void;
  /**
   * Returns the y coordinate of the vector
   */
  y(): void;
}
```

jrabek:
I am reimplementing https://github.com/rive-app/rive-wasm/blob/master/js/src/utils/registerTouchInteractions.js#L18 in a Phaser friendly way.

Related: https://github.com/rive-app/rive-wasm/issues/262

zplata:
Thanks for raising this @jrabek - we'll take a look at this and figure out what's going on. Also you're right..seems strange that it is methods for x and y. Also will take a look at changing that to properties and let you know if issues arise from that.

zplata:
This should be fixed in `1.0.86`. Just to be safe though, we'll still keep the `x` and `y` as functions on `Vec2D` but change them to properties in the next major bump of this runtime!

zplata:
Closing for now, but happy to re-open if the new version is still an issue for you!

PR #266, opened 2022-09-22
fix: Add constructor for Vec2D class and fix types

zplata:
Raised in #265 . We didn't make a constructor for the Vec2D type in the types file, so adding that, and fixing the return type for `x()` and `y()` on that class too. It was also brought up how it is strange that these are methods on the class, instead of properties, where other math-based Rive classes use properties (i.e `Mat2D`) for x and y values. Just to be on the safe side, I just put a note in that we may want to change `x` and `y` gets to a property instead of a function on `Vec2D` on the bindings in the next major bump.

One other thing as part of this PR is deleting a stale folder that isn't used in the build process, and that I think causes confusion on where the WASM binding glue code comes from. Happy to make this a separate PR though if we want to seperate concerns.


js/src/rive_advanced.mjs.d.ts
```diff
@@ -658,12 +658,13 @@ export declare class Mat2D {
  * Rive Vector class
  */
 export declare class Vec2D {
+  constructor(x: number, y: number);
   /**
    * Returns the x coordinate of the vector
    */
-  x(): void;
+  x(): number;
```
csmartdalton, line 665: 💯

csmartdalton (commented):
lgtm!

csmartdalton (approved):

PR #270, opened 2022-10-20
Fix Alignment leak!

luigi-rosso:
- Alignment now exposed as an enum so that emscripten doesn’t new it every time we get it.
- Use the same clang-format as the rest of our C++ code (gotta mino-repo this soon)
- Use address sanitizer for helping debug memory leaks in debug mode.
- Better compiler/linker flag separation between debug and release mode in our premake scripts.

### DIFF
--- a/wasm/js/renderer.js
+++ b/wasm/js/renderer.js
@@ -874,4 +874,10 @@ Rive.onRuntimeInitialized = function () {
   );
   Rive["disableFPSCounter"] = _animationCallbackHandler.disableFPSCounter;
   _animationCallbackHandler.onAfterCallbacks = flushCanvasRenderers;
+
+  Rive["cleanup"] = function () {
+    if (_rectanizer) {
+      _rectanizer.delete();
+    }
+  };
 };
--- a/wasm/src/bindings.cpp
+++ b/wasm/src/bindings.cpp
@@ -26,6 +26,8 @@
 #include "rive/shapes/path.hpp"
 #include "rive/transform_component.hpp"
 
+#include "js_alignment.hpp"
+
 #include "src/core/SkIPoint16.h"
 #include "src/gpu/GrDynamicRectanizer.h"
 
@@ -36,6 +38,7 @@
 #include <stdio.h>
 #include <string>
 #include <vector>
+#include <sanitizer/lsan_interface.h>
 
 using namespace emscripten;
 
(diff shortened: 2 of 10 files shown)


luigi-rosso:
Also bumps to latest rive-cpp master.

wasm/js/renderer.js
```diff
@@ -874,4 +874,10 @@ Rive.onRuntimeInitialized = function () {
   );
   Rive["disableFPSCounter"] = _animationCallbackHandler.disableFPSCounter;
   _animationCallbackHandler.onAfterCallbacks = flushCanvasRenderers;
+
+  Rive["cleanup"] = function () {
+    if (_rectanizer) {
+      _rectanizer.delete();
```
csmartdalton, line 880: Was this a leak?

csmartdalton (commented):
Do you think you can do the clang-format change in a follow-on PR? It makes this one hard to review.

Also, typo in the commit message. "mono" and not "mino"

wasm/js/renderer.js
luigi-rosso: No but if you run the leak test in the example app it will report leaks if you don't call this (we allocate it once like a singleton/manager and it detects that as leaked memory).

luigi-rosso:
Only reason I added the clang-format was that my vscode was trying to reformat it with default clang-format settings or not format it at all. Not sure what we were doing prior to this? Were we just not formatting any of the C++ code here?

wasm/src/js_alignment.hpp
```diff
@@ -0,0 +1,20 @@
...
+    topRight,
+    centerLeft,
+    center,
+    centerRight,
+    bottomLeft,
+    bottomCenter,
+    bottomRight
+};
+
+rive::Alignment convertAlignment(JsAlignment alignment);
```
luigi-rosso, line 19: This is the biggest change introduced.

wasm/src/bindings_skia.cpp
```diff
@@ -20,98 +20,110 @@
...
+        .function("transform", &rive::Renderer::transform, allow_raw_pointers())
+        .function("drawPath", &rive::Renderer::drawPath, allow_raw_pointers())
+        .function("clipPath", &rive::Renderer::clipPath, allow_raw_pointers())
+        .function("align",
+                  optional_override([](rive::Renderer& self,
+                                       rive::Fit fit,
+                                       JsAlignment alignment,
+                                       const rive::AABB& frame,
+                                       const rive::AABB& content) {
+                      self.align(fit, convertAlignment(alignment), frame, content);
```
luigi-rosso, line 118: We override the default align with one that accepts our enum and converts it to the matching rive::Alignment. This allows us to only expose the enum to JS.

wasm/src/bindings_c2d.cpp
```diff
@@ -20,410 +21,444 @@
...
+        .function("transform", &RendererWrapper::transform, pure_virtual(), allow_raw_pointers())
+        .function("drawPath", &RendererWrapper::drawPath, pure_virtual(), allow_raw_pointers())
+        .function("clipPath", &RendererWrapper::clipPath, pure_virtual(), allow_raw_pointers())
+        .function("align",
+                  optional_override([](rive::Renderer& self,
+                                       rive::Fit fit,
+                                       JsAlignment alignment,
+                                       const rive::AABB& frame,
+                                       const rive::AABB& content) {
+                      self.align(fit, convertAlignment(alignment), frame, content);
```
luigi-rosso, line 389: Same deal as other comment, override the default align with one that converts the simple JS Alignment enum to our actual rive::Alignment version.

csmartdalton:
> Only reason I added the clang-format was that my vscode was trying to reformat it with default clang-format settings or not format it at all. Not sure what we were doing prior to this? Were we just not formatting any of the C++ code here?

Ah ok, no worries. Your comments helped!

csmartdalton (approved):
Ship it!

csmartdalton:
> Ship it!

(After the build passes :)

wasm/src/bindings.cpp
```diff
@@ -48,440 +51,446 @@ extern rive::Factory* jsFactory();
...
+{
+    return invertedMatrix * canvasVector;
+}
 
-  int drawHeight() const { return m_Rectanizer.drawBounds().height(); }
+class DynamicRectanizer
+{
+public:
+    DynamicRectanizer(int maxAtlasSize) :
+        m_Rectanizer(SkISize::Make(1, 1),
```
luigi-rosso, line 110: This was causing an assert (it would initialize with 0,0 prior to this change) @csmartdalton

PR #272, opened 2022-10-20
Remove call to runtime cleanup

avivian:


zplata (approved):

luigi-rosso (approved):
LGTM

js/src/rive.ts
```diff
@@ -1460,8 +1460,6 @@ export class Rive {
     this.renderer.delete();
     // Delete the rive file
     this.file.delete();
-    // Cleanup anything remaining on the runtime.
-    this.runtime.cleanup();
```
luigi-rosso, line 1464: For posterity: This deletes the singleton backing Rectanizer for image mesh rendering but doesn't reset the pointer to null. So it should only be called when you know you are fully done with rive (like before checking for leaked memory in the module).

In the future we could: 
- set the rectanizer singleton to null after deleting it so anyone else that needs it will auto-regenerate it.
- just remove this cleanup call from the runtime in release builds/name it more appropriately

PR #273, opened 2022-10-20
fix: delete mat2d and vec2d created objects instanced thru WASM in listeners logic

zplata:
Making sure we clean up WASM-generated objects here


mjtalbot (approved):
lgtm i guess :D no tests is always tough, but not sure how to tests this either well

avivian (approved):

PR #276, opened 2022-11-04
add mouseover and mouseout events for pointer move

zplata:
To add more accuracy to mouse behavior, adding the `mouseover` and `mouseout` events to trigger pointer move events for Listeners. This should be triggered if for example, a mouse is in the canvas already, and then quickly moved out of the canvas, where the browser would not trigger `mousemove` across that behavior. In this case, `mouseout` would always be called.


luigi-rosso (approved):
Great!!

pa5aribu:
Hey @zplata , could this issue be related to this https://github.com/rive-app/rive-wasm/issues/277?

PR #279, opened 2022-12-08
Maint: Add eslint and prettier config

zplata:
Making linting and prettier config to standardize the format of things in our `js` folder


avivian (approved):
LGTM!

PR #281, opened 2023-01-13
fix: unload Mat2D object reference before transliterating to JS for canvas2d renderer

zplata:
**Problem**: We're noticing a memory leak that's slow-burning on the memory footprint of an app that uses `@rive-app/canvas`. Because we allow memory to grow when building WASM, it's not immediately evident unless you run an app for potentially several hours.

**tl;dr:** Unload `Mat2D` properties, which are just floats, before calling into JS for a few functions.

After turning off `'-s ALLOW_MEMORY_GROWTH=1'` in our premake config on debug mode and running against the WASM `parcel_example` app with about 20 canvases going, it'll crash the app in a few seconds. Interestingly, WASM doesn't report a memory leak using our [memory leak test function](https://github.com/rive-app/rive-wasm/blob/master/wasm/examples/parcel_example/index.js#L193). So it's not that our JS heap is growing in size, but rather something is driving memory usage up on the browser that emscripten doesn't seem to be catching.

After a bit of a chase to find the leak, it seems that any CPP binding that calls into JS (`renderer.js`) to do some calculation **with** an object reference in particular as a parameter seems to be an offending case here. This is used in a few spots with respect to `.transform()` and `.addRenderPath()` in CPP-land, which call into JS with a `const rive::Mat2D& transform` object. Pointers and primitive types don't seem to be problematic in this conversion.

Some folks have tried to raise this - but nothing super clear [here](https://github.com/emscripten-core/emscripten/issues/9930). Also, the docs don't give examples or guidance on this case so hard to tell the main issue beyond Emscripten trying to make a copy of the object before passing to JS from what I think I'm gathering.

**Solution (for now)**: What seemed to have fixed the main issue for rendering right now is instead of passing the `Mat2D` object reference to JS, just unpacking the values of it and sending it to JS instead, since they're just `float` primitives. This is a purely internal call API change, so nothing consumer-facing and shouldn't be breaking.

Next steps I think are to try and re-create this underlying reference issue in a standalone WASM example and try to raise with the Emscripten team if we can

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup emsdk
         uses: mymindstorm/setup-emsdk@v11
         with:
-          version: 3.1.20
+          version: 3.1.29
           actions-cache-folder: "emsdk-cache"
       - name: Install premake
         run: |
--- a/wasm/examples/parcel_example/index.js
+++ b/wasm/examples/parcel_example/index.js
@@ -1,9 +1,9 @@
 import "regenerator-runtime";
 // import RiveCanvas from "../../build/bin/debug/canvas_advanced_single.mjs";
-import RiveCanvas from "../../build/bin/debug/canvas_advanced.mjs";
+// import RiveCanvas from "../../build/bin/debug/canvas_advanced.mjs";
 // import RiveCanvas from "../../../js/npm/webgl_advanced_single/webgl_advanced_single.mjs";
 import { registerTouchInteractions } from "../../../js/src/utils";
-// import RiveCanvas from '../../../js/npm/canvas_advanced_single/canvas_advanced_single.mjs';
+import RiveCanvas from "../../../js/npm/canvas_advanced_single/canvas_advanced_single.mjs";
 import AvatarAnimation from "./look.riv";
 import TapeMeshAnimation from "./tape.riv";
 import BirdAnimation from "./birb.riv";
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Setup emsdk
         uses: mymindstorm/setup-emsdk@v11
         with:
-          version: 3.1.20
+          version: 3.1.29
```
mjtalbot, line 17: this intentional?

wasm/examples/parcel_example/index.js
```diff
@@ -1,9 +1,9 @@
 import "regenerator-runtime";
 // import RiveCanvas from "../../build/bin/debug/canvas_advanced_single.mjs";
```
mjtalbot, line 2: was about to comment on all the commented out lines, but i guess this is testing that the parcel examples are no longer leaking 👍

wasm/premake5.lua
```diff
@@ -128,6 +128,7 @@ do
     buildoptions {
         '-g2',
         '-fsanitize=address'
+        -- '-fsanitize=leak'
```
mjtalbot, line 131: we keeping these commented out & in here for a reason (maybe a crutch to remember we can do this?), a link to the fsanitize build options might be helpful here in that case

mjtalbot (approved):
:rubber stamp: (maybe luigi will have more input)

is there a performance difference?

.github/workflows/build.yml
zplata: Yeah - probably good to keep us up to date with emscripten updates and doesn't seem to be breaking things between these two. But also doesn't make much a difference (emscripten bumping doesn't hurt or help here)

wasm/src/bindings_c2d.cpp
```diff
@@ -77,7 +77,25 @@ class RendererWrapper : public wrapper<rive::Renderer>
...
+        call<void>("transform",
+                   transform.xx(),
+                   transform.xy(),
+                   transform.yx(),
+                   transform.yy(),
+                   transform.tx(),
+                   transform.ty());
+    }
+
+    void align(RendererWrapper& self,
```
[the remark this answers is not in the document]
zplata: its the existing `align` function but want to make sure we use this `RendererWrapper`'s  `transform` method above, rather than the existing CPP implementation in rive-cpp in `renderer.cpp` but I think we can mark this as `override` to make it clearer

.github/workflows/build.yml
mjtalbot: oh.. lol i read .19 as oppsed to .29.

csmartdalton:
Thanks for all your work tackling this @zplata ! Great findings!

luigi-rosso:
> :rubber stamp: (maybe luigi will have more input)
> 
> is there a performance difference?

I actually think performance might be slightly better like this, especially if it's newing up less objects for each call (and less calls back across the bridge to get each property).

PR #282, opened 2023-01-16
fixed iphone xr 14.5-14.6 can't display bug

ansonchen:
`const log = gl.getProgramInfoLog(program);`
In iPhone XR iOS 14.5 - 14.6 result is "↵" 
The result returned by the log is an unexpected string length.
So the main program will breaks.


zplata (approved):
Awesome, thanks for catching this!!

PR #297, opened 2023-03-17
Handle error response code when loading .riv files

avivian:
Returns early when loading a .riv if the status code is >= 400.


zplata (approved):

zplata:
We check the status now when fetching the riv and appropriately fire the loaderror event

PR #299, opened 2023-03-20
feature: support touch events for listeners

zplata:
Adding support for touch interactions (specifically to be able to support touch and drag on mobile devices)

Thinking of bumping a minor version with this. There's a slight side effect where now if you are slowly dragging your finger on the canvas, you may have scrolling disabled, due to calling `event.preventDefault()` on `touchmove` events. If you drag on the canvas rapidly, sometimes the browser scroll will still take, but not always.


js/src/utils/registerTouchInteractions.ts
```diff
@@ -10,16 +22,44 @@ export const registerTouchInteractions = ({
...
+      (event as TouchEvent).changedTouches?.length
+    ) {
+      canvasX =
+        (event as TouchEvent).changedTouches[0].clientX - boundingRect.left;
+      canvasY =
+        (event as TouchEvent).changedTouches[0].clientY - boundingRect.top;
+    } else {
+      canvasX = (event as MouseEvent).clientX - boundingRect.left;
+      canvasY = (event as MouseEvent).clientY - boundingRect.top;
+    }
```
avivian, line 62: Nitpick, but I'd extract this out into a function the returns the clientX and clientY based on the event type and then do the math to calculate the canvasX and canvasY with the result, since the logic is the same for all three cases

avivian (approved):

PR #302, opened 2023-04-02
fix: add safety checks before invoking cleanup on renderer and file on main cleanup function

zplata:
Issue brought up on Discord where a user is trying to clean up instances where a file instance may not be created yet. Adding a few null safety checks before calling `.delete()` in the cleanup function to ensure this case is handled.

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -1453,8 +1453,10 @@ export class Rive {
     this.cleanupInstances();
     // Delete the renderer
     this.renderer?.delete();
+    this.renderer = null;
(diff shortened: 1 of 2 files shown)


luigi-rosso (approved):
LGTM, but do we need to set the renderer/file to null after calling delete too? Otherwise we might try to delete the renderer/file twice if this code runs again (maybe there's a check for this higher up though).

zplata:
@luigi-rosso  Ah yes, we should do that. Especially if Emscripten is still leaving behind its JS code on the object.

Issue #309, opened 2023-04-22
Listeners only registered if state machines are playing (autostart)


FlorianPAZE:
### Description
In the example shown below, we have set autoplay to false, in order to start the playback after the "Okay" button is clicked. However, doing so, the state machine does in fact start playing - however the listeners for hover and click that we have on the state machine are not registered and don't work anymore. Setting the autplay to true works just fine.

### Provide a Repro
This example shows the issue: https://codepen.io/florianpaze/pen/abRpbaQ/479047d396a701b7c996a2a344185dbb 

#### Source `.riv`/`.rev` file
(.riv file is linked in the codepen JS)

### Expected behavior
Even if autoplay is set to false, the listeners should be properly registered, so that we can use playback controls after the fact without loosing the listener interactions.

zplata:
@FlorianPAZE - try upgrading the runtime lib to `@rive-app/canvas: 1.1.5`. This ensures that if you set `autoplay: false` and start `play()` with a state machine, Rive Listeners should be enabled.

PR #310, opened 2023-04-24
Fix: Ensure Rive listeners are setup when the play() method is called and torn down on stop()

zplata:
See [original issue](https://github.com/rive-app/rive-wasm/issues/309) for the bug.

Essentially, if someone has `autoplay: false` then later calls `.play('name-of-state-machine');` on the Rive instance later, the Rive Listeners would not be setup, and no event listeners (i.e `mouseenter`, `mousemove`, etc.) were attached to the `<canvas>` due to all the setup happening only on load/initialization of the Rive instance.

This cleans up any existing event listeners on the `<canvas>` when `play()` is called, and sets it up again accordingly, while also cleaning up the listeners on `<canvas` when `stop()` is called.


js/src/rive.ts
```diff
@@ -1066,6 +1068,7 @@ export class Rive {
     this.src = params.src;
     this.buffer = params.buffer;
     this.layout = params.layout ?? new Layout();
+    this.shouldDisableRiveListeners = !!params.shouldDisableRiveListeners;
```
mjtalbot, line 1071: 👀

js/src/rive.ts
```diff
@@ -1124,7 +1126,6 @@ export class Rive {
     artboard,
     autoplay = false,
     useOffscreenRenderer = false,
-    shouldDisableRiveListeners = false,
   }: RiveLoadParameters): void {
```
mjtalbot, line 1129: Just talking out loud here, looks like init is private, so & the constructor will take care of the "shouldDisableRiveListenders" for any situation now

js/src/rive.ts
```diff
@@ -1546,6 +1551,9 @@ export class Rive {
       return;
     }
     this.animator.stop(animationNames);
+    if (this.eventCleanup) {
+      this.eventCleanup();
```
mjtalbot, line 1555: did we never call cleanup before?

js/test/test-rive-buffers/listenerFile.js
```diff
@@ -0,0 +1,2555 @@
+export default function () {
+  return [
```
mjtalbot, line 2: not a problem with this pr, but there must be a better way?

mjtalbot (approved):
LGTM! thanks for getting tests in as well!

js/test/rive.test.ts
```diff
@@ -1124,6 +1125,59 @@ test("Mouse events are removed from canvas when reset", (done) => {
   });
 });
 
+test("Canvas has listener events attached on a Rive file with Rive Listeners", (done) => {
```
mjtalbot, line 1128: would be neat to also test listeners are gone when we've stopped. looks like thats something that's been fixed here as well?

js/src/rive.ts
zplata: Not on `stop()`. Was mainly only in the public `cleanup()` method

js/test/rive.test.ts
zplata: Oh good call! We can listen for `removeEventListener`. I think I recently discovered you can't query what event listeners are setup on a DOM element directly, so that might work

PR #319, opened 2023-06-03
docs: add changelog step into release-it workflow

zplata:
Adding `release-it` auto-changelog into github release workflow when publishing to add commits for each associated release version. Deleting the old one that referenced the deprecated `rive-js` lib


HayesGordon (approved):
LGTM - though maybe we should keep the old CHANGELOG still?

zplata:
Yeah fair point - adding a note to the old one that it's deprecated with a link to the newer one a folder up

PR #320, opened 2023-06-06
Feature: Add drawImage API to CanvasRenderer on low-level runtime and fix types

zplata:
Exposes a way to call Canvas2D's `drawImage()` API from the renderer returned from `.makeRenderer(canvas)` to allow folks to draw separate images on the canvas alongside Rive content, while making sure it stays within the deferred drawList ops.

Also fixing types for this for TypeScript defs.


csmartdalton (approved):
LGTM!

PR #321, opened 2023-06-07
Make sure Animator.add() is iterating only on own properties

Andy1210:


zplata (approved):
Thanks for your contribution @Andy1210 ! We'll get this patched up

PR #328, opened 2023-07-06
Deprecating 'unsubscribe' in favor of 'off' in JS API.

Kusmeroglu:
Making event listener registration functions consistent `on`/`off` instead of `on`/`unsubscribe`. Leaves `unsubscribe` as an alias to `off` with a deprecation warning.


zplata:
Thanks for your contribution! We got your patch of this PR applied in the latest version `v1.2.3`.

PR #337, opened 2023-11-03
Add: export ImageAsset and FontAsset types

cdelguercio:
Added so that I could cast a FileAsset to ImageAsset or FontAsset and call setRenderImage or setFont without TypeScript getting angry.


zplata:
Thanks for the PR @cdelguercio! We'll get this integrated

zplata:
This change was introduced from our upstream workflow and released in `v2.7.1` - with your commit credited here: https://github.com/rive-app/rive-wasm/commit/9e54b205091d8da141046fe485d1407a63a0b218

Thanks for the contribution!

PR #341, opened 2023-12-27
Include types declaration in webgl advanced packages

joncardasis:
## Purpose

The `@rive-app/webgl-advanced` and `@rive-app/webgl-advanced-single` packages do not contain the TS declarations in their published npm modules.

Add the generated `rive_advanced.mjs.d.ts` declarations file to the module's `files` for publishing.

### DIFF
--- a/js/npm/webgl_advanced/package.json
+++ b/js/npm/webgl_advanced/package.json
@@ -23,7 +23,8 @@
   "license": "MIT",
   "files": [
     "webgl_advanced.mjs",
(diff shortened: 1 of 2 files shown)


zplata:
Thanks for adding this! We'll get this change merged in next chance we get

zplata:
Merged this from our upstream repository with your change, and should be in v2.9.1. Feel free to open an issue if you still notice issues here

PR #343, opened 2024-01-04
add optional param for method resizeDrawingSurfaceToCanvas

Aratmany:
I have suggestions to add to the `resizeDrawingSurfaceToCanvas(devicePixelRatio)` method, an optional argument that you can pass a value to.
 
When using the `resizeDrawingSurfaceToCanvas() `=> method when resizing, it takes into account the DPR of the device and multiplies the width and height by that value. That it starts to lag on weak devices, where the difference between smaller and larger canvas resolution is not so great, but the difference in performance is noticeable.

if you use resizeDrawingSurfaceToCanvas()
Screen Shot 2024-01-04 at 14 52 54
----------------------------------

if you use resizeDrawingSurfaceToCanvas(customDPR)
Screen Shot 2024-01-04 at 14 52 15

it gives a good performance for weak devices, since the browser doesn't have to render 2x-3x canvas (the difference is almost insignificant to the eye).
customDPR (1.5) set increases the width and height of the canvas by 1.5x

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -1903,10 +1903,10 @@ export class Rive {
    * and resize the layout to match the new drawing surface afterwards.
    * Useful function for consumers to include in a window resize listener
    */
-  public resizeDrawingSurfaceToCanvas() {
+  public resizeDrawingSurfaceToCanvas(customDevicePixelRatio?: number) {
     if (this.canvas instanceof HTMLCanvasElement && !!window) {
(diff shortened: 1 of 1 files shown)


zplata:
Hey - great suggestion! I think this change makes sense for customizing and having better control over the canvas area. Will do some testing and look at merging this change in, thanks for the contribution!

zplata:
This change was merged in from our upstream repo and should be available in `v2.9.2`. Thanks again for the contribution!

Issue #360, opened 2024-07-19
Using exposed CanvasRenderingContext2D API 


DanEscamilla:
### Description
I'm having issues using the Context2D API, mainly that you cannot change the fill between multiple calls. Usually the way you interact with the Context2D api is by defining color styles and then calling draw operations. 
```
ctx.fillStyle = "blue";
ctx.fillRect(10, 10, 100, 100);
```

But it appears that the way the API works through the RiveRenderer, is to queue draw calls, but this means that the mutating of the fillStyle in a draw cycle, happens before the actual draw call is executed. This has the effect of all the draw calls using a single fillStyle. 

### Provide a Repro
```
const rive = await RiveRuntimeLoader({
    locateFile: (_) =>
      `https://unpkg.com/@rive-app/canvas-advanced@2.17.3/rive.wasm`,
  });
const riveRenderer = rive.makeRenderer(this._canvas);
const ctx = riveRenderer as (CanvasRenderingContext2D);

ctx.fillStyle = "blue";
ctx.fillRect(10, 10, 100, 100);

ctx.fillStyle = "red";
ctx.fillRect(120, 10, 100, 100);
```

#### Source `.riv`/`.rev` file
N/A
-->

### Expected behavior
In this case, both rectangles will be red, as opposed to one being blue and the other being red which is what is expected from using the native CanvasRenderingContext2D API

#### Screenshots
N/A

### Browser & Versions (please complete the following information)
- Browser: Chrome Version 126.0.6478.62 (Official Build) (arm64)
- OS: Mac OS 13.6.4 (22G513)
- NPM Version: 10.5.0
- Rive version: 2.19.4

HayesGordon:
Closing this as it's resolved in https://github.com/rive-app/rive-wasm/commit/0509735de630d3184e56fc89bae1b6f9b96522b5. Thanks for the contribution.

PR #362, opened 2024-07-23
Fix: canvas's context2d property update timing

DanEscamilla:
## Summary

This PR aims to fix the issue detailed in #360.

The issue stems from the native context2d calls being queued up to be executed later. This creates a rift between the draw calls and the setting of properties.

The proposed solution creates a closure for each property setter call and adds it to the draw queue, ensuring that the order of the draw calls and the property setters remains consistent chronologically.

### DIFF
--- a/wasm/js/renderer.js
+++ b/wasm/js/renderer.js
@@ -946,7 +946,9 @@ Module["onRuntimeInitialized"] = function () {
       },
       set(target, property, value) {
         if (property in c2dSource) {
-          c2dSource[property] = value;
+          newCanvasRenderer._drawList.push(() => {
+            c2dSource[property] = value;
+          });
           return true;
         }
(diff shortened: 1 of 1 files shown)


bodymovin (approved):
lgtm

HayesGordon:
Hi @DanEscamilla, we must merge this PR into our upstream repository. This will require a member of the Rive team to co-author this change with you. We'll let you know when that has been pushed downstream to this repository, and then we can close this issue. Thanks for the contribution!

HayesGordon:
The commit with your contribution is here @DanEscamilla : https://github.com/rive-app/rive-wasm/commit/0509735de630d3184e56fc89bae1b6f9b96522b5

We'll close this PR, and create a new release.

PR #372, opened 2024-10-30
chore: tag 2.22.0

mjtalbot:
just getting the commit in as the build with this has already landed


HayesGordon (approved):
LGTM! Thank you

PR #377, opened 2024-11-18
chore: update contributing docs for installation

lancesnider:
Minor changes to the CONTRIBUTING.md docs: 

- Install Ninja
- Change `rive-runtime` branch name from `master` to `main`
- Add troubleshooting for the Python3 error 
- Minor changes to make it more readable

CleanShot 2024-11-18 at 14 54 45@2x

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,26 +29,53 @@ For each of the variants above, there are two sets of APIs to use for rendering 
 
 ### Installation
 
-1. Clone the project down
-2. `cd` into the `js` folder from the top level of the project, and run `npm i` to install all dependencies
-3. Pull down the underlying `rive-runtime` submodule, which should be pointing to a specific commit.
+
+
+1. Install [Emscripten](https://emscripten.org/docs/getting_started/downloads.html). We build against `v3.1.43` in rive-wasm, so make sure to install and activate that version.
+2. Install [Premake5](https://premake.github.io/) and add it to your `PATH`.
+3. Install [Ninja](https://ninja-build.org/).
 
 ```sh
-# First time through if you just pulled down the project
+# Clone the Project
+git clone git@github.com:rive-app/rive-wasm.git
+
+# Install dependencies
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -29,26 +29,53 @@ For each of the variants above, there are two sets of APIs to use for rendering
...
-# When updating the submodule after the initial pull
-git submodule update --recursive
+# When updating the submodule after the initial installation
+# git submodule update --recursive
+
+# Make sure the submodule is on the main branch
+cd wasm/submodules/rive-runtime/
+git checkout main
+
+cd ../../../js
```
HayesGordon, line 58: I vote that we instead just say:
- `cd` into the js dir.
- `cd` into `wasm/submodules/rive-runtime` and run..
Instead of hardcoding the commands like this.

Or alternatively do:
```
pushd wasm/submodules/rive-runtime/
....
popd
...
pushd js
```

CONTRIBUTING.md
```diff
@@ -29,26 +29,53 @@ For each of the variants above, there are two sets of APIs to use for rendering
...
+
+Error: `./get_skia.sh: line 31: python: command not found`:
 
-6. `cd` back into the `js` folder and run `./build.sh` from your terminal/shell to build the latest WASM and builds for JS API's (high and low level) into the `npm/` folder. This may take some time, grab a coffee! This should finish with Webpack building the JS bundles for the high-level API packages (more on that below)
+```sh
+# Install Python 3.9 with Homebrew
+brew install python
+
+# Add a symlink
+ln -s  /usr/local/bin/python3.9 /usr/local/bin/python
```
HayesGordon, line 77: I don't think creating a symlink is the right call, it may interfere with the symlinks brew already makes.

I used to have an alias setup to point python to python3. But I removed that and the scripts are working for me.

@lancesnider can you check to see if it's working for you by simply having python3 installed without creating the symlink or any other preference.

PR #378, opened 2024-11-19
If there is no artboard, running cleanup() will cause an error

lancesnider:
If there is no artboard, `this.animator` will be `undefined`. If you run `cleanup()`, it will try to run `this.animator.stop(animationNames);`, which will result in the following error. 

Screenshot 2024-11-13 at 12 08 13 PM (3)

This PR makes sure `this.animator` exists before trying to call `this.animator.stop(animationNames);`.

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -2251,7 +2251,10 @@ export class Rive {
       });
       return;
     }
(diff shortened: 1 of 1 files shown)


HayesGordon (approved):
LGTM!

PR #380, opened 2024-12-10
Add deprecation notice to webgl, webgl-advanced, and webgl-single, webgl-advanced-single packages

lancesnider:
This PR adds a deprecation warning to the top of the webgl runtimes. This will also update the descriptions on the various [npm package pages](https://www.npmjs.com/package/@rive-app/webgl). 

CleanShot 2024-12-09 at 17 59 46@2x


lancesnider:
We're eventually going to deprecate webgl2 in favor of webgl so this isn't a good idea. :)

PR #410, opened 2025-10-30
feat: expose wasmBinary to allow Rive being used in environments without access to fetch and XHR

naugtur:
This change allows the end-user of canvas and react-canvas to use the previously undocumented wasmBinary config field to pass an ArrayBuffer directly.

We've arrived at this as a solution for using Rive without giving it access to fetch and XHR at all 
(it's frowned upon by browser extension review process to have code that could, even potentially, load sources from the web and execute within the extension) 

We're running Rive under LavaMoat, so we can deny it access to fetch while keeping it for ourselves. 

The option seemed very convenient, so I thought I'd offer it as a contribution.

### DIFF
--- a/js/src/rive.ts
+++ b/js/src/rive.ts
@@ -222,6 +222,8 @@ export class RuntimeLoader {
   // if embedded wasm is used then this is never used.
   private static wasmURL = `https://unpkg.com/${packageData.name}@${packageData.version}/rive.wasm`;
 
+  private static wasmBinary: ArrayBuffer;
+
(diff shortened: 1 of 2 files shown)


zplata:
Thanks @naugtur for the contribution! We picked your commit into our upstream repro which has officially landed.
You can now set `RuntimeLoader.setWasmBinary()`
https://github.com/rive-app/rive-wasm/blob/master/js/src/runtimeLoader.ts#L168

PR #416, opened 2026-04-07
fix: add new types to be released in high-level runtimes too for package.jsons

zplata:
Upstream repo will also push a change that includes new extra type files for a few high-level JS runtime package.json's. Adding it here so the next sync will not be affected. Short term fix for now


philter (approved):
LGTM!

PR #425, opened 2026-08-25
Revert "bug: check for animator before running animator.stop (#378)"

zplata:
This reverts commit 432c3b740a855b48c2d34973e2a641fd2b2fada6.

Some code from upstream conflicted with this PR that was merged a few years back. We need to revert it to unblock the upstream push to this repo. Will get this specific change added upstream.


bodymovin (approved):
LGTM

